PR #55, opened 2015-01-20
Param for Imagepath

elena1113:
Please check and merge

### DIFF
--- a/add/data/prefs/edirom-prefs.xml
+++ b/add/data/prefs/edirom-prefs.xml
@@ -25,6 +25,8 @@
         
         <entry key="application_language" value="de"/>
         <entry key="annotation_layout" value="EdiromOnline.view.window.annotationLayouts.AnnotationLayout1"/>
+        <entry key="image_prefix" value="../../../digilib/Scaler/"/>
+        
         
     </entries>
 </prefs>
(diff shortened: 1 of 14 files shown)


roewenstrunk:
The edition specific preference file is not loaded in the eutil:getPreference function. Try something like:
```xquery
declare function eutil:getPreference($key as xs:string, $edition as xs:string?) as xs:string {
     let $file := doc('../prefs/edirom-prefs.xml')
     let $projectFile := doc(edition:getPreferencesURI($edition))
     return    
        if($projectFile != null) then ($projectFile//entry[@key = $key]/string(@value))
        else ($file//entry[@key = $key]/string(@value))   
};
```
and pass the `edition` parameter to the function

roewenstrunk:
Reopening the pull request

bwbohl:
This still open?
As we decided for gitflow PRs should not be issued against master but develop instead!
This brings up the issue to make develop the default branch #57

bwbohl:
I tested this locally and the new imagePathParam works great!

nikobeer:
Tested… works!

nikobeer:
See #62. This could be closed.

nikobeer:
closed.

PR #96, opened 2017-01-18
Edition param for activeEdition

### LINKED ISSUE #90: Paramter "activeEdition"
In previous versions (like OPERA) it was possible to leave the parameter "activeEdition" in application.js empty and to define it while calling the app.js in index.html like:

`<script type="text/javascript" src="app.js?edition=/db/contents/edition-12345"/>`

This seems not to work anymore?

roewenstrunk:
If there is a query parameter „edition“ the value of this param will be
the value of me.activeEdition
(fixes #90)

### DIFF
--- a/app/Application.js
+++ b/app/Application.js
@@ -82,6 +82,10 @@ Ext.define('EdiromOnline.Application', {
             scope: this
         });
         
+        var workParam = me.getURLParameter('work');
+        if(workParam !== null)
+            me.activeWork = workParam;
+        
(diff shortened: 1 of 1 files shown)


bwbohl:
is there a way of handling activeWork, too?

roewenstrunk:
If there is a query parameter „work“ the value of this param will be
the value of me.activeWork. Please double check

nikobeer:
ok, checked.

PR #105, opened 2017-03-10
Support for HTML files displayed in an iFrame view

roewenstrunk:
cherry-picked commit f4027b61229a5530be794215ccce4bc497aa85cf for
integration in develop


roewenstrunk:
Hi Niko, as you can see in `add/data/xql/getLinkData.xql` there will be an iFrame view if you request for a document in the database with an `html` root element.

nikobeer:
Build process fails with:

[ERR] Failed to resolve dependency EdiromOnline.view.window.View for file EdiromOnline.view.window.iFrameView

Maybe some other parts of freidi-integration are missing?

bwbohl:
@nikobeer looks like you're right, a search told that there's no corresponding class defined in develop…

PR #109, opened 2017-04-03
Feature localizeNavigator

bwbohl:
Add lang localisation Options for NavigatorConfiguration.

Before this PR the Navigator would always read the first edirom:name element for Category and Item entries.
With this PR the getPreferences(‘application_lang’) will be submitted to the call getNavigatorConfig.xql and the existing edirom:name elements with appropriate @xml:lang will be processed. In case no such name is available the first name-element will be used as fallback.


bwbohl:
@nikobeer check this out ;-)

PR #122, opened 2021-02-08
Features from Beethoves-Werkstatt

bwbohl:
* search for edition and work
* alternatively get source labels from different xpath, but always return a string
* reactivate concordance navigator

all seem plausible to me


bwbohl:
closes #128

PR #160, opened 2021-12-22
Did some fixes for functions in namespace 'local' to keep compatible with eXist-db 5.3(.1)

nikobeer:
Mainly quick and dirty: move all appearances of local:getLocalized*($node) to one single eutil:getLocalizedName($node, $lang).

Further examination: eXist-db 5.3 seems to have a problem with function declarations in namespace "local" in separate xqls when theses xqls import other xqls which also contain the same function declarations. So, declaring local:getLocalizedName() in several xqls is not possible anymore…? 

It's already working in real world! Choose one Edirom from here: https://www.reger-werkausgabe.de/module-ii.html


add/data/xqm/util.xqm
```diff
@@ -34,8 +34,42 @@ import module namespace teitext="http://www.edirom.de/xquery/teitext" at "teitex
...
+        then $node/mei:name[@xml:lang = $lang]/text()
+        else $node/mei:name[1]/text()
+    )
+    else if ($node/edirom:names)
+        then (
+            if ($lang = $node/edirom:names/edirom:name/@xml:lang)
+            then $node/edirom:names/edirom:name[@xml:lang = $lang]/node()
+            else $node/edirom:names/edirom:name[1]/node()
+    )
+    else ($node)
```
bwbohl, line 69: What if the $node contains localised variants, this function would return both, wouldn't it?
Shouldn't there be a more general fallback first, e.g. something like:
```xquery
else if ($lang = $node[@xml:lang])
```

add/data/xqm/util.xqm
```diff
@@ -34,8 +34,42 @@ import module namespace teitext="http://www.edirom.de/xquery/teitext" at "teitex
...
 
 declare namespace mei="http://www.music-encoding.org/ns/mei";
+declare namespace edirom="http://www.edirom.de/ns/1.3";
+
 import module namespace functx = "http://www.functx.com" at "../xqm/functx-1.0-nodoc-2007-01.xq";
 
+(:~
+: Returns a localized string
+:
+: @param $node The node to be processed
```
bwbohl, line 44: Probably it should be clarified that not the node's but it's child nodes' values will be returned?

add/data/xqm/util.xqm
```diff
@@ -34,8 +34,42 @@ import module namespace teitext="http://www.edirom.de/xquery/teitext" at "teitex
...
+: @return The string
```
bwbohl, line 45: what string?

add/data/xqm/util.xqm
```diff
@@ -34,8 +34,42 @@ import module namespace teitext="http://www.edirom.de/xquery/teitext" at "teitex
...
+:)
```
bwbohl, line 46: feel free to add `@author`

bwbohl (commented):
What about other nodes? Maybe a general fallback that process nodes matching $lang should be provided?

bwbohl (approved):

PR #172, opened 2022-01-21
Bazga merge

bwbohl:
* adds Docker based automatic build on pullrequests
* mardown improvement in README
* introduces index.xql with CSS injection options
* adds no-break spaces to lang file de
* adds missing lang keys to en
* only show priority and category label in annotation meta if at least one value is present
* adds localization of categories and priorities
* * fixes annotation previews
* adds XQDoc to annotation.xqm and edition.xqm
* * introduces source:isSource function
* adds annotationIcons to openseadragonView
* some indentation


bwbohl:
might also close other PRs from BAZ-GA

add/data/xql/getAnnotationPreviews.xql
```diff
@@ -286,16 +286,17 @@ declare function local:getImageAreaParams($zone as element()?, $imgWidth as xs:i
 };
 
 declare function local:getItemLabel($elems as element()*) as xs:string {
-    
+    let $language := eutil:getLanguage($edition)
```
riedde, line 289: I guess this requires that lang is set in the edition(-file)? If yes, the language has to be required by the schema.

add/data/xqm/edition.xqm
```diff
@@ -111,3 +111,13 @@ declare function edition:findEdition($uri as xs:string) as xs:string {
         return 'xmldb:exist://' || document-uri($edition/root())
     )
 };
+
+(:~
+: Returns the name of the edition specified by $uri
+:
+: @param $uri The URI of the Edition's document to process
+: @return the text contents of edirom:editio/edirom:editionName
```
riedde, line 119: typo (edirom:editio**n**)

add/data/xqm/source.xqm
```diff
@@ -33,14 +34,16 @@ declare namespace mei="http://www.music-encoding.org/ns/mei";
...
 : @param $uri The URI of the document
 : @return Is work or not
 :)
 declare function source:isSource($uri as xs:string) as xs:boolean {
-    
-    exists(doc($uri)//mei:mei) and exists(doc($uri)//mei:source)
+    let $doc := eutil:getDoc($uri)
+    return
+        exists($doc//mei:mei) and exists($doc//mei:source) (:mei2 and ?3 :)
+        or ($doc//mei:mei/@meiversion = "4.0.1" and exists($doc//mei:manifestation[@singleton='true'])) (:mei4 :)
```
riedde, line 46: What about `$doc//mei:mei/@meiversion/starts-with(., "4.0")` in face somebody uses 4.0.0?

riedde, line 46: `@singleton='true'` what does that mean?

add/index.xql
```diff
@@ -0,0 +1,78 @@
...
+        <!-- **ACE** -->
+        <script src="resources/js/ace/ace.js" type="text/javascript" charset="utf-8"/>
+        <script src="resources/js/ace/mode-xml.js" type="text/javascript" charset="utf-8"/>
+        
+        <link rel="stylesheet" href="resources/EdiromOnline-all.css"/>
+        <!-- TODO if prefs css then include here -->
+        {
+            if ($preferences//entry[@key = "additional_css_path" and @value != ''])
+            then
+                <link rel="stylesheet" href="{string-join((request:get-context-path(), 'rest', substring-after($preferences//entry[@key = 'additional_css_path']/@value, 'xmldb:exist:///')), '/')}"/>
```
riedde, line 66: Does this require to insert the value as `db/apps/` instead of `/db/pps/`? I think the other perf entries use leading slash

riedde (commented):
I had a look and added just some comments and questions. I also tried to understand the js-script changes but I'm not qualified for that ;)

riedde:
@bwbohl There was a minor merge conflict. I resolved it. After that the checks startet, but failed. Just for your interest.

add/data/xqm/source.xqm
bwbohl: I like the idea

bwbohl: manifestation singleton is borrowed from FRBRoo and describes a manifestation that exclusively cannot have other items e.g. an autograph. The idea behind is to not over-emphasise on dividing metadata between item and manifestation in such cases

add/data/xqm/edition.xqm
bwbohl: fixed

add/data/xql/getAnnotationPreviews.xql
bwbohl: yes or in a cookie('edirom-language') edirom edition ns 1.3 does require `edition/languages/language` but yes it does not require `xml:lang` to be anything different than ""

bwbohl: just opened https://github.com/Edirom/edirom-edition-schema/issues/8

add/data/xqm/source.xqm
bwbohl: will hard code 4.0.0 and 4.0.1 as there probably will be no other MEI v4.*.*

bwbohl: created an issue at https://github.com/music-encoding/music-encoding/issues/906

add/index.xql
bwbohl: don't get it, could you please elaborate?

bwbohl: if you mean as value in the prefs file? the query assumes that the path might be prefixed with `xmldb:exist:///` and removes it from the return string

bwbohl: The query assumes that the css path might be prefixed with `xmldb:exist:///` and removes it just in case, if it's not there path will be returned and yes, then you are right. But I always assumed all paths in prefs should be prefixed wit `xmldb:exist:///`

riedde: thank you for clarifying!

bwbohl:
tested locally with current EditionExample, seems to work

bwbohl:
tested locally and works with current EditionExample

PR #174, opened 2022-02-02
Improve readme

bwbohl:


bwbohl:
added some info on local testing

PR #175, opened 2022-02-02
Fix get concordance

bwbohl:
actually this PR reworks util.xqm

* the reason for getConcordance to fail was in util:getLocalizedName returning non-normalised $node if no explicit rule was defined for a node. This breaks the returned concordance-JSON due to the contained line-breaks
* adding fn:normalize-space fixed it

Moreover this PR does some Linton of the eutil module


bwbohl:
accidentally did some necessary workflow updates in this PR

PR #180, opened 2022-03-08
locale files: replace most whitespaces with nbsp

### LINKED ISSUE #133: use no break spaces in langVars of UI mouseover titles
Depending on the browser the whitespaces in the mouseover labels will result in line breaks. the results range from unreadable content to blocked UI elements. This might be fixed by using no-break spaces in the langVars

bwbohl:
closes #133


nikobeer:
Like! :-)

PR #182, opened 2022-03-08
Fix desktop home btn label

bwbohl:
* introduce HomeBtnLabel langvar
* fix work combobox title population to accomodate MEI v3 and MEI v4


riedde (approved):
fine!

riedde:
I have tested this and I also added a little css quick fix for the vertical align of the home button label. It is again hard coded...
From my view: can be merged!

Issue #75, opened 2015-10-13
Selection page in Verovio-View


elena1113:
Selection of page in Number-Field functions not correctly

bwbohl:
deprecated, there is no entry field for page-numbers in the verovio view anymore

Issue #74, opened 2015-10-13
Add Progress-bar to Verovio-View


elena1113:
Add progress-bar to three views in Verovio-View during data load.

bwbohl:
now there is an icon indicating verovio execution before display of notation --> closing

PR #193, opened 2022-03-09
Update verovio

riedde:
The implemented verovio toolkit is outdated. I have thrown out the code and introduced the verovio-app. 

ToDo/Bugs:
- [x] The height of the iframe is not equal to the window height. I was not able to fix that.


riedde:
closes #74 
closes #75

riedde:
@roewenstrunk Could you have a look at the window height bug? Thanks!

kepper:
Just had a first look, a CSS rule `.renderingViewContent {
  height: 100%;
}` already gets us closer. We probably need to substract the height of the page turn buttons (`calc(100% - 60px)` or some such)…

bwbohl:
@riedde & @roewenstrunk could you please resolve the conflicts! And investigate how this does relate to #215 ?

riedde:
I had a quick look on the conflict. We have to discuss this in our upcoming community meeting, because the canges belong to different ways of using Verovio and one excludes the other. Issue #215 should be solved with this update no matter the way we choose

bwbohl:
@riedde: @roewenstrunk already merged some mods concerning verovio. Would be nice if you two could sort this out

riedde:
- The issues #74 and #75 must be evaluated, because they might be obsolete in some may. It depends on how we are continueing.
- This issue definitely closes #215, the rendering is now working.

I further resolved the conflicts between develop and this PR. The decition I made was to use the verovio app and not, as before, the verovio toolkit with an selfmade navigation. I did so because the adjustments @roewenstrunk did match the integrated verovio version of 2.7.2, but the current release is 3.9.0 and I do not know if this will raise other issues! The verovio-app was easy to integrate, has its own navigation and is working fine so far.

- At the moment my code gets the verovio engine by https-request but the code of the verovio-app is already integrated. A next step would be to also integrate all online dependencies.

However, we should discuss that at the next meeting, because this will impact some other functionalitites of the view.

bwbohl:
please postpone this update for a later release

riedde:
Even if the verivioApp is quite nice styled and has a lot of build-in functions, it would be too difficult to get back the interactions with Edirom. The way it is implemented now (revision done in 1.0.0-beta) is the better way to go!

Issue #186, opened 2022-03-08
add language parameter to AJAX calls


bwbohl:
As with the editionId always being submitted in AJAX calls it would be good to also generally submit the current application language as AJAX parameter. (AJAXController.js)

PR #198, opened 2022-03-18
controller: add lang attribute to url

### LINKED ISSUE #197: work switch requires lang parameter in url
Switching between works is only possible if the 'lang' parameter is set (e.g., http://localhost:8080/exist/apps/Edirom-Online/index.html?lang=de).
If this is not set, after choosing another work the navigator is showing a load spinner and doesn't react.

riedde:
quick fix for switchiing works, closes #197

### DIFF
--- a/add/controller.xql
+++ b/add/controller.xql
@@ -15,7 +15,7 @@ if ($exist:path eq "") then
 else if ($exist:path eq "/") then
     (: redirect root path to index.html :)
     <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
-        <redirect url="index.html?lang=de"/>
+        <redirect url="index.html?lang={eutil:getLanguage(request:get-parameter("edition", ""))}"/>
     </dispatch>
 else if ($exist:path eq "/index.html") then
     (: forward index.html to index.xql :)
--- a/add/controller.xql
+++ b/add/controller.xql
@@ -1,5 +1,7 @@
 xquery version "1.0";
 
+import module namespace eutil = "http://www.edirom.de/xquery/util" at "data/xqm/util.xqm";
+
 declare variable $exist:path external;
 declare variable $exist:resource external;
 
--- a/add/controller.xql
+++ b/add/controller.xql
@@ -13,7 +13,7 @@ if ($exist:path eq "") then
 else if ($exist:path eq "/") then
     (: redirect root path to index.html :)
     <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
-        <redirect url="index.html"/>
+        <redirect url="index.html?lang=de"/>
     </dispatch>
 else if ($exist:path eq "/index.html") then
     (: forward index.html to index.xql :)


add/controller.xql
```diff
@@ -13,7 +13,7 @@ if ($exist:path eq "") then
 else if ($exist:path eq "/") then
     (: redirect root path to index.html :)
     <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
-        <redirect url="index.html"/>
+        <redirect url="index.html?lang=de"/>
```
bwbohl, line 16: thanks for spotting the problem! Nevertheless I think, the value of the lang variable should not be hard-coded to `de` but instead retrieved from preferences or even better from the current app or user agent lang. What do you think?

bwbohl (changes requested):

add/controller.xql
riedde: That was a quick fix because of the upcoming release. But you're right it shouldn't be hard-coded. What about introducing a hidden parameter an a function that fills in the browser lang? As a fallback english would be better I think. I opened an issue concerning switching languages some days ago.
I'll provide a proposal on the handling of the lang parameter

riedde: I had a quick look and found out that we do not need to implement very much. Edirom uses a cookie named `edirom-language` [see here](https://github.com/Edirom/Edirom-Online/blob/4e374d2fd403baa682ae47eafbbdd04847a30fa8/add/data/xqm/util.xqm#L191).
So we just have to fill in the right value or make the controller to read this correctly.

riedde: just to have it documented: It seems to me that the error could also occur, if the edirom edition file contains only naked `<name>` elements that do not have `@xml:lang` attributes. In that case the default (english) could not be loaded?
When testing with the edition example edirom always returns the english values, even if 'edirom-prefs', 'project-prefs' and 'edition-file-lang' are set to `de`. This might be a small bug with some dependencies

bwbohl: > I had a quick look and found out that we do not need to implement very much. Edirom uses a cookie named `edirom-language` [see here](https://github.com/Edirom/Edirom-Online/blob/4e374d2fd403baa682ae47eafbbdd04847a30fa8/add/data/xqm/util.xqm#L191). So we just have to fill in the right value or make the controller to read this correctly.

same direction I was investing!
Controller just would have to import the `util.xqm` module and then could use the `eutil:getLanguage` function to retrieve the correct lang value. Probably we should also include lang in all AJAX requests as mentioned in #186

bwbohl: Will you do this or shall I?

bwbohl: The edition should probably alsoo be retrieved in the controller in the same way the edirom retrieves the first eidtion.xml it finds in the database if no parameter is set in the url.

riedde:
I updated the code. now it works. The parameter is shown in the url, but it is filled in dynamically.

bwbohl:
works here, too, thanks!
One thing we could think about is adding it to the index.xql call in order to keep the uri neat and clean, e.g.:

```xquery
else if ($exist:path eq "/index.html") then
    (: forward index.html to index.xql :)
    <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
        <forward url="index.xql">
            <add-parameter name="lang" value="{eutil:getLanguage(request:get-parameter("edition", ""))}"/>
        </forward>
    </dispatch>
    ```

bwbohl:
probably controller.xql would also be a good place in call `edition:findEdition` or even `edition:getUris` and then redirect to a modal if multiple editions are found!

bwbohl:
we should go with your solution for now and improve later ;-)

bwbohl (approved):

PR #204, opened 2022-03-28
Fix broken navigator

riedde:
If there are more (text) values of titles than one (occurs in MEI 4.0.1), then ne Navigator brakes. To avoid that the incoming values are joined.


riedde:
Please test this with your data. on my machine something was broken and I couldn't test it. The code looks good so there shouldn't be a cause why not.

Issue #218, opened 2022-05-04
getMeasuresOnPage.xql sometimes seems to return too many measures


bwbohl:
cf. our one test dataset

PR #221, opened 2022-05-04
fix filter for measures on page

riedde:
Before contains() was used but this produces too many matches if the IDs contains floating numbers (multiple matches). Now the exact id must occur for a match.

### DIFF
--- a/add/data/xql/getMeasuresOnPage.xql
+++ b/add/data/xql/getMeasuresOnPage.xql
@@ -45,7 +45,7 @@ declare option exist:serialize "method=text media-type=text/plain omit-xml-decla
 declare function local:getMeasures($mei as node(), $surface as node()) as xs:string* {
 
     for $zone in $surface/mei:zone[@type='measure']
-    let $measures := $mei//mei:measure[@facs = concat('#', $zone/@xml:id)]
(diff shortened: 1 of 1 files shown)


riedde:
closes #218

bwbohl:
I will test this

add/data/xql/getMeasuresOnPage.xql
```diff
@@ -44,7 +44,7 @@ declare option exist:serialize "method=text media-type=text/plain omit-xml-decla
 declare function local:getMeasures($mei as node(), $surface as node()) as xs:string* {
 
     for $zone in $surface/mei:zone[@type='measure']
-    let $measures := $mei//mei:measure[contains(@facs, concat('#', $zone/@xml:id))]
+    let $measures := $mei//mei:measure[@facs = concat('#', $zone/@xml:id)]
```
bwbohl, line 47: please consider the case that a `@facs` can contain multiple references, e.g. by tokenizing the attribute value and testing against the set of tokens

bwbohl (changes requested):

add/data/xql/getMeasuresOnPage.xql
riedde: done

PR #224, opened 2022-05-12
Fix tei footnotes

bwbohl:
* reconceptualise transformation of TEI texts to HTML by relying primarily on TEI Stylesheets and extracting from the resulting HTML
* fix some CSS and SCSS conflicts


bwbohl:
open issues:

- [ ]  footnote heading
- [ ] footnote links and backlinks

riedde (commented):
Reviewed and seems to be okay. Thank you for your work! I'll test this and then approve my review.

riedde:
@bwbohl the footnote links and backlinks are working. When backlinking the target is a bit out of the window so a bit padding would be nice, but from my view it works!

bwbohl:
I think we can live with this for the moment ;-)

PR #225, opened 2022-05-15
extend getParts for all MEI verisons

riedde:
I just added a short XPath to locate performance resources encoded after 2012.


riedde:
Tested. Could be merged before the release.

bwbohl (approved):

PR #234, opened 2022-06-01
Proxy for accessing ressources in the database

roewenstrunk:
The proxy can be used to get access to documents and files in the database. The parameter `ressourcePath` is used to point to the file (e.g. `/db/apps/doc/resources/images/existdb-web.svg`). The parameter `contentType` is used to give the content type for the HTTP response (e.g. `application/xml`, `image/png`, `text/css;%20charset=UTF-8`, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types).

### DIFF
--- a/add/data/xql/proxy.xql
+++ b/add/data/xql/proxy.xql
@@ -1,0 +1,43 @@
+xquery version "3.0";
+(:
+  Edirom Online
+  Copyright (C) 2022 The Edirom Project
+  http://www.edirom.de
+
+  Edirom Online is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Edirom Online is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


riedde (commented):
Couldn't test it because on my local machine it was not broken. We should test this in an online database with proxy.
@roewenstrunk Does this work in a standalone mode OR do we have to include this xql somewhere else?

peterstadler:
Hmm, if this PR is supposed to be a workaround for CSS injection with disabled REST interface, I think this has the same potential security risks as the generic REST interface. With knowledge of the db structure an attacker can access/download every file.

bwbohl:
@peterstadler thanks for the valuable advice!
Would it be more sensible then, to calculate the proper frontend URI of a referenced `additional_css_path` based on exist:context and $editionUri? E.g. similar to:

https://github.com/Edirom/Edirom-Online/blob/e3f4be1386651a1c791f6903346ce58a3c11f537/add/data/xslt/teiBody2HTML.xsl#L199-L204

with `$context-path` being request:get-context-path()?

peterstadler:
As proposed in #241 I think we'll need some convention on where the Edirom root controller accepts requests for custom resources, e.g. `$edirom-root$/custom-resources/$edition-id$` (with "custom-resources" being this hard coded convention). 
Under this virtual path the controller might then serve resources from arbitrary collections within the db.

PR #242, opened 2022-06-07
generalize metadata view to handle all `meiHead` child elements

bwbohl:


riedde (commented):
looks good. Thanks for the work. I'll test, approve and merge it

Issue #241, opened 2022-06-07
ability to inject custom CSS and Javascript


peterstadler:
IFF I understand correctly, at present only CSS (not JS) injection is possible via eXist's REST interface.

My proposal is to create an Edirom way to inject both JS and CSS, not relying on the REST interface since this should IMHO be disabled for production use. It'll need 

* a user preference setting (per edition) that will depict the path to a CSS and a JS folder in the data package
* a function to insert every CSS (and JS) file found within those folders into the Edirom HTML
* a Edirom controller rule to forward everything under e.g. `custom-resources/$edition-id$/css` to the respective collection in the data package

If this seems a viable approach I'm happy to work on this and send in a PR!

bwbohl:
Created a pull request to avoid using the rest API for accessing the injected CSS file. In terms of injecting JS I wonder if you could give me some use-cases? We already do have a plugin mechanism for overriding XSLT, XQuery and JS files or functions of the Edirom-Online (unfortunately undocumented beyond the use in different edition projects). What else would you imagine?

bwbohl:
Cinsidering that a productive environment might not allow requests such as proposed in #244 I think you're absolutely right @peterstadler that a sound and solid concept working on the basis of eXist-db controller architecture has to be developed.

bwbohl:
Did some digging and testing and apparently the plugin mechanism relies on the rest interface, too…
So yeah we should make use of the exist-db controller mechanism to come up with a general-purpose solution.

bwbohl:
With #244 and #172 it should be possible to add custom CSS via edition preference `additional_css_path`. I know it’s not perfect, but it might do the trick for the moment. I think a more profound remodelling could be tackled after [1.0.0-Release](https://github.com/Edirom/Edirom-Online/milestone/14). @peterstadler ?

peterstadler:
I'm absolutely ok with whatever you decide(d) at the community meeting. I have only organizational questions: should we keep the issue open and remove it from the [1.0.0-Release](https://github.com/Edirom/Edirom-Online/milestone/14)?
Or should we close it and open a new one for the "profound remodelling"?

bwbohl:
Pushing this to 1.1.0 although one might even argue this is a major change ;-)

PR #244, opened 2022-06-07
get edition frontend uri to fetch injected CSS

bwbohl:
partially addresses #241 
might close #234


riedde (commented):
reviewed, but not tested yet. Should be tested on a server that uses a proxy.

riedde (approved):
Thank you for your proposal. It's now also working on my machine!

PR #252, opened 2022-06-14
Update getLinkTarget.xql

### LINKED ISSUE #250: metadata view is default view for all content types
Before v1.0.0-beta.1 each content type had its default view. Since the release, the metadata view is the global default view. This should be fixed.

bwbohl:
reenable setting of default views closes #250


riedde (approved):
Thank you for figuring this out

PR #254, opened 2022-06-15
Fix annotation previews

bwbohl:
- [x] Still some problems with JSON decoding
- [x] image size does not fit in preview area


roewenstrunk:
Can you, @bwbohl, explain a little more precise what has to be done here?

riedde (approved):
Very nice! Thank you. It's tested and fine!
I found an ongoing issue but we should merge this now and handle the other separately

PR #255, opened 2022-06-15
Feature: improve concordance measure handling

riedde:
- calculated measure labes if multirest
- sorting measure numbers containing letters (92, 92a, 93)
- maybe something else...


roewenstrunk (approved):
Everything works as expected

PR #263, opened 2022-06-22
Start documents as uris in prefs file

roewenstrunk:
You can now put a set of uris in the prefs file with the key `start_documents_uri`. These uris will be opened when Edirom starts. If you have uris as parameters in the url, the preference will be ignored.


riedde (approved):
Thank you for this. It might be very helpful!

PR #267, opened 2022-06-22
rewrite queries to improve performance and output proper JSON

peterstadler:
While debugging via monex I found out that the functions `local:findAnnotations` and `local:getMeasures` were some bottlenecks since they were not able to employ the indexes properly.  

The main improvement (hack?) of this PR is to do a two-step query. A first predicate (which gets evaluated first in eXist) uses a simple `contains()` to narrow down the result set and the second predicate then picks the proper results from the smaller result set. 

A second modification introduced here is to return proper JSON (i.e. the mime type "application/json") which seems to work fine :)

### DIFF
--- a/add/data/xql/getAnnotationsOnPage.xql
+++ b/add/data/xql/getAnnotationsOnPage.xql
@@ -40,8 +40,8 @@ declare namespace ft="http://exist-db.org/xquery/lucene";
 declare namespace xmldb="http://exist-db.org/xquery/xmldb";
 declare namespace output = "http://www.w3.org/2010/xslt-xquery-serialization";
 
-declare option output:method "text";
-declare option output:media-type "text/plain";
+declare option output:method "json";
+declare option output:media-type "application/json";
 
 (: Returns a JSON array of annotations
  : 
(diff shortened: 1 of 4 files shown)


add/data/xql/getAnnotationsOnPage.xql
```diff
@@ -40,8 +40,8 @@ declare namespace ft="http://exist-db.org/xquery/lucene";
 declare namespace xmldb="http://exist-db.org/xquery/xmldb";
 declare namespace output = "http://www.w3.org/2010/xslt-xquery-serialization";
 
-declare option output:method "text";
-declare option output:media-type "text/plain";
+declare option output:method "json";
+declare option output:media-type "application/json";
```
riedde, line 44: We had some problems with the json output because there were some places in the js code that liked to have plain text to convert it to json. But this works on my machine. I'm surprised, but happy with it!

riedde (approved):
Thank you @peterstadler for that speedup. It is incredibly fast on my loca machine!

riedde:
@peterstadler @roewenstrunk @bwbohl 
I fixed the occurance of too many annots on facsimiles and at the same time speeded up by modifying one line in https://github.com/Edirom/Edirom-Online/commit/86f98dcc1114576880aa4d3e05674cd43dbe14de ;-)

add/data/xql/getAnnotationsOnPage.xql
bwbohl: +1

PR #274, opened 2022-06-24
Feature: annotation placement options

bwbohl:


riedde:
after #275 was approved and merged, this can be closed I think.

PR #275, opened 2022-06-24
Fix annotation placement

roewenstrunk:
Since I couldn't figure out what was happening in #274, I opened another branch with a solution for better positioning of annotations in measures.


roewenstrunk:
I will now check and fix the annotation filter for priorities and categories

riedde (approved):
Thank you @roewenstrunk for your work.

PR #278, opened 2022-06-25
Feature annotations in rendering

roewenstrunk:
First version of annotations in verovio. The annotations are always visible by now and don't react on the filter for priorities and categories.
Tooltips and links to the annotations work, but need some more work.


riedde (approved):
Thank you for your work! As you said, there is a little work to do, but this will be an other task.

riedde:
Just for documentation purpose: the CSS for anotations in the facsimile view has no effects to the rendering view (at the moemnt).

PR #287, opened 2022-11-23
Footnotes in mei annot text

nikobeer:
(cherry picked from commit 23b070d050c6e6ff13609d9c775f867208438905)

# Conflicts:
#	add/data/xslt/meiP2html.xsl

### DIFF
--- a/add/data/xslt/meiP2html.xsl
+++ b/add/data/xslt/meiP2html.xsl
@@ -64,7 +64,7 @@
         <xsl:variable name="footnoteCount">
             <xsl:value-of select="functx:index-of-deep-equal-node($footnotes/mei:annot, .)"></xsl:value-of>
         </xsl:variable>
-        <span class="sup"><xsl:value-of select="$footnoteCount"/></span>
+        <span class="superscript"><xsl:value-of select="$footnoteCount"/></span>
     </xsl:template>
     <xsl:template match="mei:ref">
         <span class="ref">
--- a/add/data/xslt/meiP2html.xsl
+++ b/add/data/xslt/meiP2html.xsl
@@ -28,7 +28,7 @@
             <div style="display:flex; color: grey; font-size: smaller; margin-bottom:0.5rem;">
                 <span class="superscript">
                     <xsl:value-of select="$counter"></xsl:value-of>
-                </div>
+                </span>
                 <div style="display: inline; margin-left: .25rem;">
                     <xsl:copy-of select="$content"></xsl:copy-of>
                 </div>
--- a/add/data/xslt/meiP2html.xsl
+++ b/add/data/xslt/meiP2html.xsl
@@ -1,8 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
-	xmlns:xs="http://www.w3.org/2001/XMLSchema" 
-	xmlns:mei="http://www.music-encoding.org/ns/mei" 
-	xmlns:xd="http://www.oxygenxml.com/ns/doc/xsl" exclude-result-prefixes="xs xd mei" version="2.0">
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:mei="http://www.music-encoding.org/ns/mei" xmlns:xd="http://www.oxygenxml.com/ns/doc/xsl" xmlns:functx="http://www.functx.com" exclude-result-prefixes="xs xd" version="2.0">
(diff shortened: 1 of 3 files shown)


riedde (commented):
Thank you @nikobeer for this feature. I just added the css class "sup" you used in L. 67 to the todo.css

riedde (approved):
I did a quick test. Works fine!

bwbohl:
Great, thanks!

Concerning the `sup` CSS class, it would be better to instantly add it to the respective scss file from the beginning. Also, textVieContent uses `.superscript` for that purpose which is already being allied in the annotation content box.

resources/css/todo.css
```diff
@@ -202,6 +202,11 @@ span.italic {
     font-style: italic;
 }
 
+span.sup {
+    font-size: smaller;
+    vertical-align: super;
+}
+
```
bwbohl, line 209:
```suggestion
```

Remove as superfluous.

add/data/xslt/meiP2html.xsl
```diff
@@ -26,6 +60,12 @@
             <xsl:apply-templates select="* | text()"/>
         </p>
     </xsl:template>
+    <xsl:template match="mei:annot[@type = 'note' and @place = 'foot']">
+        <xsl:variable name="footnoteCount">
+            <xsl:value-of select="functx:index-of-deep-equal-node($footnotes/mei:annot, .)"></xsl:value-of>
+        </xsl:variable>
+        <span class="sup"><xsl:value-of select="$footnoteCount"/></span>
```
bwbohl, line 67:
```suggestion
        <span class="superscript"><xsl:value-of select="$footnoteCount"/></span>
```

use existing superscript CSS class.

bwbohl (changes requested):

add/data/xslt/meiP2html.xsl
```diff
@@ -15,8 +12,45 @@
...
+    <xsl:variable name="footnotesBlock">
+        <xsl:for-each select="$footnotes/mei:annot">
+            <xsl:variable name="counter">
+                <xsl:value-of select="position()"></xsl:value-of>
+            </xsl:variable>
+            <xsl:variable name="content">
+                <xsl:apply-templates/>
+            </xsl:variable>
+            <div style="display:flex; color: grey; font-size: smaller; margin-bottom:0.5rem;">
+                <div style="display: inline; vertical-align: super; font-size: smaller;">
```
bwbohl, line 29: I just realised the direct CSS formatting of the footnote number. Couldn't we use `.superscript` here, too? Or may you define a class `.footnote-index` or similar?

riedde: We'll take the `.superscript` class. (Decision at meeting)

riedde (dismissed):
I added the superscript to sass so it can handle both numbering inside an annotView content

packages/eoTheme/sass/etc/annotation.scss
```diff
@@ -80,6 +80,11 @@ $content_baseFont_size: 14px;
 		    font-size: 1.1em;
 		}
 		
+		.superscript{
```
bwbohl, line 83: superscript was already defined for textVieContent and consequently worked for annotations, too (at least for spans). This more general rule does not break anything but we should consider reworking the SASS files at some point.

add/data/xslt/meiP2html.xsl
bwbohl:
```suggestion
                <span class="superscript">
```

be aware that the closing tag has to be changed, too!

add/data/xslt/meiP2html.xsl
```diff
@@ -15,8 +12,45 @@
...
+                    <xsl:value-of select="$counter"></xsl:value-of>
+                </div>
```
bwbohl, line 31:
```suggestion
                </span>
```

add/data/xslt/meiP2html.xsl
```diff
@@ -15,8 +12,45 @@
...
+                <div style="display: inline; margin-left: .25rem;">
```
bwbohl, line 32: Is there a need for this to be a div with `display: inline`?

packages/eoTheme/sass/etc/annotation.scss
riedde: In this case it is nested to annotations. But yes, we should work on this again!

add/data/xslt/meiP2html.xsl
riedde: I'm not sure. I didn't remove it because I thought that divs could have another behaviour than spans

bwbohl: ok, hopefully it works with the span above :-/

bwbohl (approved):

PR #288, opened 2022-11-25
Fix/rendering goto measure

bwbohl:
Adds go-to-measure functionality in VerovioView.
Includes format and indent of some resources.


bwbohl:
Any chance I get positive review from someone?

riedde (approved):
Works :)

PR #296, opened 2022-12-22
Fix: multiple editions handling

riedde:
The overview was working, but the selected editions were not loading. This fix makes it possible to handle multiple Edirom editions in one Edirom-Online instance.

### DIFF
--- a/add/data/xqm/edition.xqm
+++ b/add/data/xqm/edition.xqm
@@ -100,7 +100,7 @@ declare function edition:getPreferencesURI($uri as xs:string) as xs:string {
 : Only succeeds if the supplied id is the @xml:id of a edirom:edition element in '/db/apps'.
 : If $editionID is the empty string, returns the URI of the first edition found in '/db/apps'.
 :
-: @param $editionID The ID of the Edition's document to process
+: @param $editionID The '@xml:id' of the edirom:edition document to process
 : @return The URI of the Edition file
 :)
 declare function edition:findEdition($editionID as xs:string) as xs:string {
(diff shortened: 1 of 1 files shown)


bwbohl:
Wondering why this worked in my local test instance?

riedde:
@bwbohl I was wondering, why this was not working in my local test instance. However, does the fix also run on your instance?

add/data/xqm/edition.xqm
```diff
@@ -96,19 +96,19 @@ declare function edition:getPreferencesURI($uri as xs:string) as xs:string {
 };
 
 (:~
-: Returns the URI of the first found Edition
+: Returns the URI of the called Edition. Fallback: first found Edition
```
bwbohl, line 99:
```suggestion
: Returns the URI of the edition specified by the submitted $editionID parameter.
: Only succeeds if the supplied id is the @xml:id of a edirom:edition element in '/db/apps'.
: If $editionID is the empty string, returns the URI of the first edition found in '/db/apps'.
```

add/data/xqm/edition.xqm
```diff
@@ -96,19 +96,19 @@ declare function edition:getPreferencesURI($uri as xs:string) as xs:string {
...
 :
-: @param $uri The URI of the Edition's document to process
-: @return The URI
+: @param $editionID The ID of the Edition's document to process
```
bwbohl, line 101:
```suggestion
: @param $editionID The '@xml:id' of the edirom:edition document to process
```

bwbohl (approved):

Issue #293, opened 2022-12-20
ftr: part default selection by encoding


riedde:
Add a conditional to getParts.xql for setting parts visibility to default false, if encoded so.
Usecase: Multiple handwritten parts (e.g., 3 times Violin I) but only one should be shown by default.

bwbohl:
What about switching to https://github.com/music-addressability/ema/issues for such links in Edirom Online?

riedde:
If I remember right, I've done this within #294 and #300

PR #300, opened 2022-12-22
Feature: Improve handling of part labels

riedde:
- The part labels in the measureBasedView are accessed by a function.
- The label in the view and the label in the filter menu are identical now.
- The label is translated by the dictionary if the value is available (else the original value will be returned)
- if a list with a numbering is used, e.g., `violin.i` or `trumpet.2` the numbering will be treated as `Violin I` or `Trumpet 2`

### DIFF
--- a/add/data/xql/getParts.xql
+++ b/add/data/xql/getParts.xql
@@ -23,7 +23,7 @@ declare namespace request="http://exist-db.org/xquery/request";
 declare namespace mei="http://www.music-encoding.org/ns/mei";
 declare namespace xmldb="http://exist-db.org/xquery/xmldb";
 
-import module namespace eutil="http://www.edirom.de/xquery/util" at "/db/apps/Edirom-Online/data/xqm/util.xqm";
+import module namespace eutil="http://www.edirom.de/xquery/util" at "../xqm/util.xqm";
 
 declare option exist:serialize "method=text media-type=text/plain omit-xml-declaration=yes";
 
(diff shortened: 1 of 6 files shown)


riedde:
closes #293

riedde:
There seems to be a bug coming up with 7e8b7041da9fd604d8605210275f2ef34a0442f4

When activating the invisible parts, they will be added at the end and not at the expected position.

Expected:
- Violin I (1 Ex.)
- Violin I (2 Ex.)
- ...
- Doublebass

Result:
- Violin I (1 Ex.)
- ...
- Doublebass
- Violin I (2 Ex.)

riedde:
Addition: The parts appear in the order they are activated manually

add/data/xql/getParts.xql
```diff
@@ -21,15 +21,23 @@ xquery version "1.0";
 
 declare namespace request="http://exist-db.org/xquery/request";
 declare namespace mei="http://www.music-encoding.org/ns/mei";
-
 declare namespace xmldb="http://exist-db.org/xquery/xmldb";
 
+import module namespace eutil="http://www.edirom.de/xquery/util" at "/db/apps/Edirom-Online/data/xqm/util.xqm";
```
bwbohl, line 26: please fix URI

bwbohl (changes requested):

bwbohl (approved):

peterstadler (approved):

PR #303, opened 2022-12-29
Fix: annotationList and annotationPreview

riedde:
The annotation list didn't show the source sigla because of an outdated xpath (MEI3 and earlier). This PR provides an update for MEI4 and MEI5-dev. Further the annotation preview was broken because of the `source:isSource()` function. It was just checking manuscripts or sources in MEI3 or older. Now it checks also prints and also later versions, e.g. MEI5-dev using a regex.

Note: Since a manuscript requires `<manifestation singleton="true"/>`, I added analogously that a print requires `<item>`.


bwbohl:
while I really deem these changes good ones, I think maybe we should come up with a central function to get the MEI version and use that in every place that we need a distinction between MEI versions.

bwbohl (approved):

PR #310, opened 2023-04-11
Fix: adjust return values to function signatures

### LINKED ISSUE #309: issue with function return types
There seems to be a changed behavior with eXist versions > 6.0.1 that now `()/string(@foo)` returns an empty-sequence, not a string. Saxon also processes it that way so I believe it to be in alignment with the specs ;)

Various functions in edition.xqm rely on the old buggy  behavior, e.g. https://github.com/Edirom/Edirom-Online/blob/ab873e092af1aea89c68d2b1135b856448476f5b/add/data/xqm/edition.xqm#L93-L96 

That is, when `doc($uri)//edirom:preferences` evaluates to the empty sequence, the whole expression will return an empty sequence which is not what the function return type says, hence an error will be raised.

This happens in the Edirom when no resolvable `$uri` (= $uri equals the empty string) is passed to the function – which is another issue because Edirom relies on some buggy eXist behavior, see https://github.com/eXist-db/exist/issues/3898

Since this affects a lot of functions and I do not have the needed overview, I can only suggest to 
a) wrap the `fn:doc` function in a custom function that checks `doc-available` first (and/or wraps it in a try/catch)
b) replace occurrences of `some/xpath/string(@someAttribute)`  with `some/xpath/@someAttribute => string()`

peterstadler:
This is my try to fix #309. I basically replaced `/string(@attribute)` with `/@attribute => string()` to make sure the return type is `xs:string` even when the leading XPath expression returns the empty sequence.

NB, no error checking was introduced, so if e.g., (for any reason) there are multiple `edirom:language` with the same `@xml:lang`, the function `edition:getLanguageFileURI` will break.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Thumbs.db
 /tmp
 *.xpr
 local.properties
+.idea/
--- a/add/data/xqm/edition.xqm
+++ b/add/data/xqm/edition.xqm
@@ -69,7 +69,7 @@ declare function edition:getUris() as xs:string* {
 :)
 declare function edition:getWorkUris($uri as xs:string) as xs:string* {
     
-    doc($uri)//edirom:work/string(@xlink:href)
+    doc($uri)//edirom:work/@xlink:href ! string(.)
 };
 
 (:~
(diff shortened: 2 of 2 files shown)


bwbohl:
Maybe we should do a systematic search across all XQueries if such cases still appear

peterstadler:
> Maybe we should do a systematic search across all XQueries if such cases still appear

Yes, imho the Edirom would benefit a lot from systematic refactoring but I dare to say that this is out of scope of the current issue.

bwbohl (approved):

Issue #185, opened 2022-03-08
replace all Ext.Ajax.request calls with window.doAJAXRequest calls


bwbohl:
Edirom Online has a own AJAXController defined in AJAXController.js, the benefits of using it are e.g. that the edition [editionId] is always submitted as parameter in those request calls. Although considered the ‘new’ way of doing AJAX requests in Edirom Online this has been in place for years but never thoroughly implemented on project scale.

Consequently `Ext.Ajax.request` calls should be considered deprecated and replaced everywhere.

PR #308, opened 2023-02-17
Fix/307 missing lang param

bwbohl:
* closes #307 
* adds new functions and perspectives on language handling

Issue #104, opened 2017-02-17
Error handling for AJAX requests


roewenstrunk:
We need error handling for the global AJAX request function in `controller/AJAXController.js`. This is in general related to #15

PR #323, opened 2023-05-31
Integrate fix/ajax-requests into develop

### LINKED ISSUE #186: add language parameter to AJAX calls

roewenstrunk:
This PR is related to #185 and #186 

closes #186


bwbohl (changes requested):
After testing a clean build and a build after merging to develop I still get the following error loading the Edirom Online:

```
app.js:51852 Uncaught TypeError: Cannot read properties of undefined (reading 'data/xql/getEditionURI.xql')
    at constructor.getPreference (app.js:51852:22)
    at method (app.js:3269:17)
    at constructor.doAJAXRequest (app.js:51786:25)
    at method (app.js:3269:17)
    at constructor.launch (app.js:56922:10)
    at constructor.onBeforeLaunch (app.js:31602:13)
    at constructor (app.js:31563:6)
    at new constructor (app.js:4677:31)
    at app.js:6485:36
    at app.js:7450:22
getPreference @ app.js:51852
method @ app.js:3269
doAJAXRequest @ app.js:51786
method @ app.js:3269
launch @ app.js:56922
onBeforeLaunch @ app.js:31602
constructor @ app.js:31563
constructor @ app.js:4677
(anonymous) @ app.js:6485
(anonymous) @ app.js:7450
fire @ app.js:7417
readyEvent.fire @ app.js:7574
fireReadyEvent @ app.js:7620
(anonymous) @ app.js:3284
setTimeout (async)
defer @ app.js:3333
fireDocReady @ app.js:7610
onReadyEvent @ app.js:7600
```

Apparently something is going wrong, ~~might be the return type of getEditionURI.xql~~.[EDIT]

bwbohl:
Can #104 also be considered fixed with this PR?

app/controller/PreferenceController.js
```diff
@@ -30,18 +30,18 @@ Ext.define('EdiromOnline.controller.PreferenceController', {
...
             },
-            scope: this
-        });
+            Ext.bind(function(response){
+                this.setPreferences(Ext.JSON.decode(response.responseText));
+            }, this),
+            2, // retries
+            false // async
+        );
     },
```
bwbohl, line 45: The Edirom Online fails with

```
app.js:51858 Uncaught TypeError: Cannot read properties of undefined (reading 'data/xql/getPreferences.xql'
```

unless this stays the old way or the override check is removed from the AJAXController

bwbohl (changes requested):
Either we remove the override possibility from the AJAXController OR we stick to the old Ext.Ajax.request for initializing the preferences OR we somehow avoid that when getPreferences.xql is called.

bwbohl:
After that is fixed it still fails, when tryinig to `getPreference` as long as the preferences aren't available yet.

This might fix it:
```js
    getPreference: function (key, lax) {
        var me = this;

        (async () => {
            console.log("waiting for preferences");
            while (me.preferences === undefined) // define the condition as you like
            await new Promise(resolve => setTimeout(resolve, 1000));
            console.log("preferences defined");
            if (! me.preferences[key] && lax) {
                return null;
            }
            if (key == 'application_language') {
                var lang = me.getURLParameter('lang');
                if (lang) {
                    return lang;
                } else {
                    return 'en';
                }
            }
            if (me.preferences[key] == '' && key == 'image_prefix') {
                return '';
            }
            if (! me.preferences[key]) {
                Ext.Error.raise({
                    msg: 'No preference found with this key', key: key, level: 'warn'
                });
                return null;
            }
            return me.preferences[key];
        })();
    }
```

bwbohl:
After that the language problem kicks in for getLanguageString…

so maybe we should test with a merged #308 ?

bwbohl:
pinging @roewenstrunk

bwbohl:
pinging @roewenstrunk

bwbohl:
Attempt to boil it down to some understandable language:

1. When Edirom-Online loads without any further parameter in the call, it does not know the active edition and thus does not know where to find the edition preferences.
2. The AJAXController has an override variable that comes from the old plugin mechanism. The idea was that for any JS or XQuery file in the Edirom-Online code a custom file could be injected by defining the file path of the replaced file as preference `@key` and supplying the path to the injected file as value, e.g.:
   ```xml
   <entry key="plugin_annotView" value="../../../exist/rest/db/$CONTENTS_FOLDER$/$EDITION_ID$/$ANY_PATH$/$FILE_NAME$.js"/>
 ```
 3. This results in a check for a 'plugin' when trying to load the preferences, which cannot succeed because the definition of a plugin would be in the preferences and of course injecting another preferences file this way would be complete nonsense.

Theoretically, possible solutions could be to:

1. Avoid override check when fetching preferences
2. Catching an error on the override variable when trying to fetch an override defined in the preferences fails
3. getting rid of the Plug-In mechanism which actually isn't state of the art – relying on rest – and might pose major security concerns.

bwbohl (approved):

app/controller/PreferenceController.js
bwbohl: added an override exclusion mechanism

PR #333, opened 2023-06-01
Sync main and develop

bwbohl:
Somehow there are differences between main and develop where there should be none… This PR brings that in sync again.


riedde (approved):
I only detected whitespace changes.

PR #350, opened 2024-02-05
improve: renderingView fitting to window

bwbohl:
* revert some changes from #342 
* move style definitions to eoTheme sass


riedde (approved):
reviewed and tested

PR #354, opened 2024-02-05
Fix: Annotation tooltip (layout)

bwbohl:


riedde:
@bwbohl Yes, this feature should be activated. I've tested it and noticed, that there is a problem on the layout and dimensions of the annot-tooltip. Maybe we can fix that also in this PR

bwbohl:
I think, this was in the wake of reactivating everything instead of simply commenting it out in the sourcecode.

bwbohl:
I think this should go it could be hidden via CSS if someone does not want it in their edition.
A future perspective would be to add som annotation tooltip layouts to choose from, similar to the annotation layouts.

riedde:
I'm not sure why I did this adjustment, but it seems to be a bigger issue. Maybe opening an Issue and closing this PR draft?

bwbohl:
Considering that reactivating the metaBox will pile up more additional work than we want for the upvomin 1.0.0 release, I'd opt for pushing this to 1.1.0, too. Any objections?

daniel-jettka:
moved to https://github.com/Edirom/Edirom-Online-Frontend/issues/43

closing here.

PR #363, opened 2024-02-07
Fix MEI version check in source:isSource function

bwbohl:
* makes the regEx match MEI 5.0 version numbers
* makes the regex check more strict `^…&``
* fixes and adds documentation of source:isSource

### DIFF
--- a/add/data/xqm/source.xqm
+++ b/add/data/xqm/source.xqm
@@ -44,7 +44,7 @@ import module namespace eutil="http://www.edirom.de/xquery/util" at "../xqm/util
 :)
 declare function source:isSource($uri as xs:string) as xs:boolean {
     let $doc := eutil:getDoc($uri)
-    let $meiVersion4To5Regex := '^[4-5](\.\d)+(-dev)?(\+(anyStart|basic|CMN|Mensural|Neumes))?$'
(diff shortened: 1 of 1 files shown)


add/data/xqm/source.xqm
```diff
@@ -34,20 +34,29 @@ declare namespace mei="http://www.music-encoding.org/ns/mei";
...
 : @param $uri The URI of the document
-: @return Is work or not
+:
+: @return xs:boolean indicating whether the document referenced by @param $uri
+: is considered a music source or not.
 :)
 declare function source:isSource($uri as xs:string) as xs:boolean {
     let $doc := eutil:getDoc($uri)
-    let $meiVersionRegex := '(([4-9])|(\d+[0-9]))\.\d+\.\d+(-dev)?'
+    let $meiVersion4To5Regex := '^[4-5](\.\d)+(-dev)?(\+(anyStart|basic|CMN|Mensural|Neumes))?$'
```
riedde, line 47:
```suggestion
    let $meiVersion4To5Regex := '^[4-5](\.\d){1,2}(-dev)?(\+(anyStart|basic|CMN|Mensural|Neumes))?$'
```

Just a variation, but maybe not necessary. The version numbers have a second or third level. We could make this explicit with `{1,2}` instead of `+` (see suggestion). But maybe a bit oversized.

riedde (dismissed):
Great regex! Hopefully the MEI community doesn't change everything for version 6 so it doesn't get more complex ;-)

PR #390, opened 2024-06-26
Fix: annotationOverlay due to invalid XQL

### LINKED ISSUE #389: pageBasedView: annotation preview fails to load
This apparently is due to a bug in the XQuery fetching the annotation preview. The error reported is:

```xml
<exception>
  <path>/db/apps/Edirom-Online/data/xql/getAnnotation.xql</path>
  <message>err:XPST0003 Syntax error within user defined function local:calculatePreviewsForTip: unexpected token: then [at line 284, column 17]</message>
</exception>
```

The corresponding section the XQuery:

https://github.com/Edirom/Edirom-Online/blob/78d4d323a5f3c817a41b97bcbc0896e75b1b3d09/add/data/xql/getAnnotation.xql#L283-L287

bwbohl:
closes #389

### DIFF
--- a/add/data/xql/getAnnotation.xql
+++ b/add/data/xql/getAnnotation.xql
@@ -280,7 +280,7 @@ declare function local:calculatePreviewsForTip($participants as xs:string*) {
     for $zone in $zones
     let $e := $elems[substring(@facs, 2) = $zone/@xml:id][1]
 
-    let $e := +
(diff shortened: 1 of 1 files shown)


riedde (approved):
Sorry that it took so long! I approve it without testing. It's only deleting a fragment that shouldn't be there!

Issue #403, opened 2024-08-14
[ftr] add code of conduct


krHERO:
**Is your feature request related to a problem? Please describe.**
Despite github recommendations for [community standards](https://github.com/Edirom/Edirom-Online/community) edirom-online does not have a CODE_OF_CONDUCT yet.

**Describe the solution you'd like**
We should follow the [suggestion](https://opensource.guide/code-of-conduct/) of github and "Wherever you can, use prior art. The [Contributor Covenant](https://contributor-covenant.org/) is a drop-in code of conduct that is used by over 40,000 open source projects, including Kubernetes, Rails, and Swift."
Because of this, the [contributor covenant](https://www.contributor-covenant.org/) will be chosen.

krHERO:
although we still have a minor discussion at #405, the PR added a code of conduct and this issue will be closed.

PR #405, opened 2024-08-15
add code of conduct

krHERO:
this PR adds a code of conduct file to the edirom-online repository

Refs #403

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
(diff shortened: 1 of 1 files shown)


bwbohl (approved):
As this code of conduct speaks of community leaders we probably should clearly articulate who such a leader is?

krHERO:
> As this code of conduct speaks of community leaders we probably should clearly articulate who such a leader is?

thanks for the review and your request. i suggest to talk about that at the next community meeting 4.9.2024.

daniel-jettka (approved):
Looks good, Thank you.

krHERO:
> > As this code of conduct speaks of community leaders we probably should clearly articulate who such a leader is?
> 
> thanks for the review and your request. i suggest to talk about that at the next community meeting 4.9.2024.

we tried to figure out, if and how a pre-defined group of edirom developers in github could be linked, but failed. 
we want to replace the term "leader" with the group of the current edirom developer team (and planned to ideally set a link to this team), but without writing single names of persons. 
regarding [github docs on team visibility](https://docs.github.com/en/organizations/organizing-members-into-teams/about-teams#team-visibility) that seems to be impossible: "People who are not members of the organization cannot view any teams"

I suggest to use the term "edirom developers" to replace it with leaders (the info-mail address is already placed). all current developers are following the code of conduct so far and are able to monitor it and i assume we would not accept future edirom developers, that are not following the rules of the code of conduct. so we should be prepared well enough with responsible persons if something is happening from the outside regarding comments on issues etc. 

any thoughts or alternatives about this suggestion?

Issue #370, opened 2024-03-20
TextFacsimileSplitView: TEI Facsimile not loading


riedde:
The TEI facsimile view was fixed at #356 #339
Maybe a side-effect

riedde:
@bwbohl seems that the FacsimileView is renewed now, but the similar code for the TextFacsimileSplitView needs the same. I tried it, but my skills on this are not too bad to manage.

krHERO:
@bwbohl and @nikobeer could you please provide some more information about this isse and what already was happening. could especially some information regarding openseadragon be helpful?

obertsalome:
The views of text and facsimile are working separately, but the split view of both text and facsimile is not working because the facsimile is not displayed:
![FacsimileOnly](https://github.com/Edirom/Edirom-Online/assets/43742700/8c2bed33-86f4-4b4d-b268-7e530cff56b8)
![TextOnly](https://github.com/Edirom/Edirom-Online/assets/43742700/43deb98d-099d-4e81-bf00-1fe9280dbbf8)
![SplitView](https://github.com/Edirom/Edirom-Online/assets/43742700/8bce996c-1eee-46c3-a807-fe6baedb4a7c)

riedde:
Side comment concerning tei-facsimileView: There is an Issue in the TaMuDi-Project to have the svg overlays also in TEI-facsimiles and not only in the (mei) sourceView. When the facsimile view is touched it can be thought about activating the overlays there. I'm willing to help.

Diginaut:
After comparing @bwbohl changes in `FacsimileViewer.js` for a similar error maybe `view.setImageSet(pages);` in line 66 of `TextFacsimileSplitView.js` has to be changed to `me.pagesLoaded(pages, view);`?

musicEnfanthen:
@obertsalome Is this facsimile also referenced via a IIIF-Uri like in the example in #339 ?

bwbohl:
@Diginaut did you test this?

Als the probelm ist that the split view is missing a statement about openSeadragon as viewer

musicEnfanthen:
@Diginaut I guess you will need to add the definition of pagesLoaded at the bottom as well, like in https://github.com/Edirom/Edirom-Online/blob/9121944b02f86ec178e9547b9334e0ee29f9a4c1/app/controller/window/text/FacsimileView.js#L89-L91

musicEnfanthen:
[...]

PR #413, opened 2024-09-05
checkout repository recursively in Github action 

peterstadler:

@obertsalome  and I were testing the current build artifact and realized that the TEI Stylesheets were missing in the EOL xar which caused the Edirom to fail miserably.  Introducing `submodules: 'recursive'` will hopefully fix this – or is there any reason to exclude the submodules from the build here?

Refs #370

## How Has This Been Tested?
Not at all

## Types of changes
- [x] New feature (non-breaking change which adds functionality)


bwbohl:
Many thanks for spotting this!

bwbohl (approved):

Issue #362, opened 2024-02-06
XQuery: replace all exist:serialize options with output:* options


bwbohl:

peterstadler:
I'm volunteering to do this :)

PR #414, opened 2024-09-05
replace proprietary `exist:serialize` options

peterstadler:

This PR replaces all occurrences of the proprietary `exist:serialize` options with the respective serialization options from the [XSLT and XQuery Serialization 3.1 standard](https://www.w3.org/TR/xslt-xquery-serialization-31/).

I tried to stick to the task but couldn't help to adjust some namespace declarations when I spotted some missing declarations or some superfluous ones.

Also, when the output method was "text" I omitted the "omit-xml-declaration" directive since this is not applicable to the text output method.

Refs #362


bwbohl:
Wow, fantastic, I hope to find the time for a look!

bwbohl:
Spotted two remianing exist:serialize statements in:

 * add/index.xql
 * add/controller.xql
 
 Other than that, great work!

bwbohl (approved):

peterstadler:
> Spotted two remianing exist:serialize statements in:
> 
> * add/index.xql
> * add/controller.xql
> 
> Other than that, great work!

Good catch, thank you! Will fix this in a second!

PR #418, opened 2024-09-06
replace proprietary exist:serialize options

peterstadler:
leftovers from https://github.com/Edirom/Edirom-Online/pull/414 spotted by @bwbohl

Refs #362 

### To discuss

The current output method  of `add/controller.xql` ("method=xhtml media-type=application/xhtml+html") seems incorrect to me. In fact, it returns a `<dispatch>` element so it should be considered "method=xml media-type=application/xml". Looking at various eXist apps (eXide, Monex, etc.), those never state the output method explicitly, so I assume they are relying on the default "xml". Only the Dashboard app has something different, yet I consider this an error?
https://github.com/eXist-db/dashboard/blob/35ca2c9d5efc6ac27b28a6bfdb4f499bb622a2d8/controller.xql#L5-L7

Nevertheless, I'm not 100% sure so I copied over the existing media type and output method and put it here for discussion.

### DIFF
--- a/add/controller.xql
+++ b/add/controller.xql
@@ -1,14 +1,27 @@
 xquery version "3.1";
 
+(: IMPORTS ================================================================= :)
+
 import module namespace edition="http://www.edirom.de/xquery/edition" at "data/xqm/edition.xqm";
 import module namespace eutil = "http://www.edirom.de/xquery/util" at "data/xqm/util.xqm";
+
+(: NAMESPACE DECLARATIONS ================================================== :)
+
+declare namespace exist="http://exist.sourceforge.net/NS/exist";
+declare namespace output = "http://www.w3.org/2010/xslt-xquery-serialization";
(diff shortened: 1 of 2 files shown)


bwbohl (commented):
thx for fixing the leftovers!

I think you’re completely right that the serialization options are not appropriate for the controller.xql

peterstadler:
Hi @bwbohl , why are you adding all these merge commits (that will apparently be included in the history)? 
I don't think they are necessary here since there were no conflicts, right? If you want a clean history, I recommend to rebase the branch. That will require a force push so should better be done by the PR opener (if he/she is not unresponsive).

peterstadler:
> The controller serialisation question is clear from what you stated above. Please don't hesitate to remove the serialisation.
Done!

bwbohl (approved):
many thanks

PR #424, opened 2024-09-11
show cases -> showcases

bwbohl:
typo correction


peterstadler (changes requested):
I agree with the corrected spelling yet there is at least another section heading (line 21) that would need fixing

bwbohl:
thx for spotting

Issue #337, opened 2023-06-13
getLinkTarget.xql fails if window title not found for mei:work


bwbohl:
https://github.com/Edirom/Edirom-Online/blob/914ccfd7dc9f3f10cbbf9ab9640980e764dc2f04/add/data/xql/getLinkTarget.xql#L170

e.g. in MEI 4.0.1 `work` does no longer contain `titleStmt`

bwbohl:
https://github.com/Edirom/Edirom-Online/blob/914ccfd7dc9f3f10cbbf9ab9640980e764dc2f04/add/data/xql/getLinkTarget.xql#L172

fails for MEI 4 upward because `mei:workDesc` was replaced with `mei:workList`

peterstadler:
Is the function failing by throwing an error or is it returning "unknown" where you would expect some title text?

riedde:
I have tested this with an MEI 5 file. I hade to delete a lot of titles before nothing was returned, but I was not able to force a fail. At the end an emptry string was retuned.

Since @bwbohl mentioned #118 here: Special characters and hard linebreaks have been a problem since I can remember. The linebreak issue was simply fixed by normalizing the space.
But also with a title containing the apostrophe I am not able to reproduce this case. @bwbohl Could you provide me a title-Container that I can copy&paste into my test set?

bwbohl:
The current implelemtation

https://github.com/Edirom/Edirom-Online/blob/d3f4ce45220f9d30c42d632eec43b7c9fc5a461b/add/data/xql/getLinkTarget.xql#L150-L189

riedde:
The normalisation already happens in `eutil:getLocalizedTitle()`

bwbohl:
hmm… please test #103 it fails in my local instance

riedde:
I'll have an eye on that!

Issue #118, opened 2017-09-28
fullwidht apostrophe  – Window fails to load


bwbohl:
When a fullwidth Apostrophe »'« (U+FF07) is part of the value being evaluated as window title the resource . fails to load

peterstadler:
can you elaborate a little bit, please? Which XQuery(?) function fails?

bwbohl:
Apparently this is quite old and probably also connected to #337

By the time I reported these content windows in the Edirom Online GUI failed to load contents and loaded as empty blank windows. All this was because the window title had some unexpected character, e.g. apostrophe (tick) or newline character. Probably normalizing space or even the proper serialization of requests might fix those issues.

riedde:
The same for me, I also can't reproduce this. @bwbohl would you share a file with me that produces the fail? Maybe the problem is caused by a different incident.

peterstadler:
I really do think this issue is outdated and fixed by proper serialization. I propose to close the issue if we do not come up with a reproducible test case in the next week

bwbohl:
It

> The same for me, I also can't reproduce this. @bwbohl would you share a file with me that produces the fail? Maybe the problem is caused by a different incident.

The issue is so 2017 ;-). I really don't think I still have the file lying around. So, I'm +1 for closing this as it is not reproducible.

bwbohl:
@thx for all your efforts @riedde & @peterstadler

Issue #103, opened 2017-02-08
fails to load Source with linebreak in <title>


bwbohl:
e.g.

```xml
<titleStmt><?prüfen?>
    <title type="abbreviated">
        <identifier>P1-PA<rend rend="sup">1</rend>
        </identifier> – Autographe Partitur</title>
    <title type="main">Autographe Partitur</title>
    <title type="sub">Manifestation</title>
</titleStmt>
```

bwbohl:
As this issue is quite old, we should first try to reproduce it before we take further steps.

bwbohl:
cannot reproduce, closing

bwbohl:
As mentioned in #337 this had to do with the text not being normalised. Currently it works for `source objects` but fails for , e.g.`text objects`.

@riedde  Just paste https://github.com/Edirom/Edirom-Online/issues/103#issue-206211166 as titleStmt in a TEI file and you get a HTTP 4000 Bad Request, the content window in the Edirom stays blank:

![Image](https://github.com/user-attachments/assets/5f8118a0-6aad-45c3-9c54-b1b379e067a7)

```xml
<?xml version="1.0" ?><exception><path>/db/apps/Edirom-Online/data/xql/getLinkTarget.xql</path><message>exerr:ERROR The actual return type does not match the sequence type declared in the function's signature: eutil:getLocalizedTitle(node(), xs:string?) xs:string. Expected cardinality: exactly one, got 2. [at line 105, column 41, source: /db/apps/Edirom-Online/data/xql/getLinkTarget.xql]
In function:
	eutil:getLocalizedTitle(node(), xs:string?) [181:10:/db/apps/Edirom-Online/data/xqm/util.xqm]
	local:getWindowTitle(node()+, xs:string) [311:18:/db/apps/Edirom-Online/data/xql/getLinkTarget.xql]</message></exception>
```

peterstadler:
Yes, there's a lot of problematic `text()` in `eutil:getLocalizedTitle#2`. 
`text()` will return text nodes, i.e. for an element like 
```xml
$foo := <foo>text1<!-- comment --> text2 </foo>
```

`$foo/text()` will return two nodes.
So, `$foo => normalize-space()` is always a good option or at least `$foo => string`.

peterstadler:
That said, `eutil:getLocalizedTitle#2` could definitely be refactored.

[...]

PR #426, opened 2024-09-11
Fixing: Work title "not found", "fail" etc.

riedde:
I introduced a function for joining strings and normalizing space because this combination is used several times and not consistent.

Refs #103 #337 #118

## How Has This Been Tested?
Test have been done with the dataset of Baumann-Digital (Cantata, see https://doi.org/10.5281/zenodo.10072485) and WeGA-Klarinettenquintett. As no fail could be reproduced, the code snippets from the refered issues have been implemented and the upcoming errors have been fixed.

## Types of changes
 
## Checklist

### DIFF
--- a/add/data/xqm/util.xqm
+++ b/add/data/xqm/util.xqm
@@ -91,7 +91,7 @@ declare function eutil:getLocalizedName($node, $lang) {
         if($node/edirom:names) then
             ($name)
         else
-            (eutil:joinedAndNormalize($name))
+            (eutil:joinAndNormalize($name))
 
 };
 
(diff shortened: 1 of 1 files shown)


add/data/xqm/util.xqm
```diff
@@ -399,3 +399,14 @@ declare function eutil:request-lang-preferred-iso639() as xs:string {
             "en"
 
 };
+
+(:~
+ : Returns one joined and normalized string
+ :
+ : @param $string The string(s) to be processed
```
peterstadler, line 406:
```suggestion
 : @param $strings The string(s) to be processed
```

add/data/xqm/util.xqm
```diff
@@ -108,25 +108,25 @@ declare function eutil:getLocalizedTitle($node as node(), $lang as xs:string?) a
   
     let $titleMEI :=
         if ($lang != '' and $lang = $node/mei:title/@xml:lang and not($node/mei:title/mei:titlePart)) then
-            ($node/mei:title[@xml:lang = $lang]//text() => string-join() => normalize-space())
+            (eutil:joindAndNormalize($node/mei:title[@xml:lang = $lang]//text(), ''))
```
peterstadler, line 111: I consider all these node selectors superfluous since the title node will be atomized anyway when thrown into some functions that works on strings. 
```suggestion
            (eutil:joindAndNormalize($node/mei:title[@xml:lang = $lang], ''))
```

peterstadler (changes requested):
I have not tested in detail yet but only looked through the code. Except from one typo I would only complain about all the extra node selectors and would recommend to introduce yet another convenience function, i.e. a 1-arity version of `eutil:joindAndNormalize` following the model of `string-join`: https://www.w3.org/TR/xpath-functions-31/#func-string-join

bwbohl:
> I have not tested in detail yet but only looked through the code. Except from one typo I would only complain about all the extra node selectors and would recommend to introduce yet another convenience function, i.e. a 1-arity version of `eutil:joindAndNormalize` following the model of `string-join`: https://www.w3.org/TR/xpath-functions-31/#func-string-join

Indeed,when at it a 1-arity version would be a good idea!

riedde:
> I have not tested in detail yet but only looked through the code. Except from one typo I would only complain about all the extra node selectors and would recommend to introduce yet another convenience function, i.e. a 1-arity version of `eutil:joindAndNormalize` following the model of `string-join`: https://www.w3.org/TR/xpath-functions-31/#func-string-join

I did some approvements. Not tested yet, because of bad internet connection. I'll tell when this is done!

riedde:
Works on my machine ;-)

riedde:
@peterstadler I understand your intention and have switched the functions name.

peterstadler (approved):
Thanks for all the fixes! I have tested briefly on my machine with the WeGA Klarinettenquintett and it did not fail ;)

peterstadler:
My suggestion for merging here would be to squash and merge since that'd get rid of several unnecessary commits (including the merge from dev).  But if the policy is to keep everything I'm fine with that too …

peterstadler:
I see that "squash and merge" is not enabled for this repo. That's probably on purpose?

bwbohl:
with this merged, can we consider the referenced issues (#103, #118, #337) closed?

riedde:
I'll check that and if the issues are outdated I'll close them.

Issue #335, opened 2023-06-02
Gently fail Edirom Online if no Edition is available in database


bwbohl:
If no edition is installed in the database the Edirom Online fails ungently with:

```xml
<exception>
<path>/db/apps/Edirom-Online/index.xql</path>
<message>err:FODC0005 exerr:ERROR Error. [at line 95, column 5] In function: edition:getPreferencesURI(xs:string) [9:38:/db/apps/Edirom-Online/data/xqm/edition.xqm]</message>
</exception>
```

This should be caught and a statement, e.g., "no edition available in the database", should be printed.

bwbohl:
The current design of `index.xql` always assumes that there is at least one Edirom Edition installed in the dartabase. Consequently `index.xql` will always produce an error when trying to fetch the name of that single edition via preferences, as no preferences can be found…

This logic probably should be handled via eXist-db `controller.xql` and forward to different templates.

peterstadler:
@riedde created a landing page for multiple editions in 5c54b4727f44e32b44c06e8238ab59e28ff0721d.

So I assume only 1. and 4. still need to be implemented, right?

riedde:
Yes, I'm not sure, if 1. is already possible. I know that @nikobeer is into this. Concerning 4, I see two was: Creating an "error-page" for this case or including the edition-example repo as a submodule so the edirom is never empty. For the last one we would need a technique (e.g., an option) to deactivate this example if not needed.
Just an idea, but I think an error handler would be most sufficent.

peterstadler:
implemented by #429, hence closing here

PR #429, opened 2024-09-18
Introducing "error page" if no edition exists (prevents failing)

riedde:
HTML page added for this case. It is analog to the page of multiple editions, but with a message that no edition exists. If no edition is found the controller will lead to this page. This prevents a total fail of the app and looks much better ;-)

Refs #335 

## How Has This Been Tested?
On local machine with docker and exist-db 6.2.0.

## Types of changes

## Checklist


add/index.xql
```diff
@@ -46,10 +46,54 @@ let $comment := comment{"
 "
 }
 
+let $eoEditionFileNone := <html xmlns="http://www.w3.org/1999/xhtml">
```
peterstadler, line 49: there's already a lot of HTML in this file and a lot of duplicated code which makes it hard to maintain. My recommendation would be to move this HTML to an extra folder (e.g. "templates") and try to create some HTML snippets that could be reused by other HTML files via xinclude.

add/index.xql
```diff
@@ -20,8 +20,8 @@ declare option output:omit-xml-declaration "yes";
 (: VARIABLE DECLARATIONS =================================================== :)
 
 declare variable $edition := request:get-parameter("edition", "");
-declare variable $editionUri := edition:findEdition($edition);
-declare variable $preferences := doc(edition:getPreferencesURI($editionUri));
+declare variable $editionUri := if($edition) then edition:findEdition($edition) else();
```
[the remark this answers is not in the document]
riedde: We can do that, but this will take some more time, because of the cascade. Functions like edition:getPreferencesURI#1 are not accepting an empty sequence. So they have to be reworked too.

add/index.xql
riedde: I'm absolutely with you! My first try was to define a html-page and redirect to there (as an error handler does). The problem was, that the fail couldn't prevent that way, because it arises in the module edition.xqm. But it should be possible by a little more work

add/index.xql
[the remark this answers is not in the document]
peterstadler: Well, this is not longer true for `edition:getPreferencesURI#1` because it got changed recently ;)
https://github.com/Edirom/Edirom-Online/blob/7899824ebda18697a3c48b6e6a1ddd90d6365826/add/data/xqm/edition.xqm#L130

But I agree, there are other functions that would need some fixing. Still, I believe that the other functions are wrong in this regard.

add/index.xql
peterstadler: Actually, I wouldn't consider this an error (that would return some HTTP error code). 
For me, it's a valid status – the application is running but there's no data to be displayed. That information need to be displayed to the user.

add/index.xql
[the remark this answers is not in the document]
riedde: I agree, so I will have to investigate the weight of this work.

add/index.xql
riedde: I'm with you, but explained it very bad. I don't want to return an error, but use something like the error handler of exist's controller.

krHERO:
consensus community meeting 9.10.2024: catch it in index.xql (no templating) with text like "no edition file found"

add/index.xql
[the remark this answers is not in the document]
riedde: Not for v1.0.0

add/index.xql
riedde: The easiest way was to reduce redundancy and simplification. Using the templating module is oversized here.

peterstadler (approved):
Just tested locally and I can confirm that the EOL now smoothly handles 0 to 2 editions

PR #434, opened 2024-10-09
shell script to start sencha

riedde:
This PR contains a very small shell script to run a docker containter with sencha for building the edirom

## How Has This Been Tested?
I removed the absolute paths and added public valiables. It works on my machine but should work on every machine (Mac).

## Types of changes

## Checklist
- My change requires a change to the documentation.


riedde:
@krHERO To be discussed with the developer team: We can also integrate this script into the build process.

daniel-jettka:
> @krHERO To be discussed with the developer team: We can also integrate this script into the build process.

I like the idea to integrate this into the build process. 
[build.sh](https://github.com/Edirom/Edirom-Online/blob/aadb46b7ce62c9523c09d6e393b5d5688654659d/build.sh) could test if sencha and ant are available locally and if not then start the build process using the container.

Small addition starts the build process in the container directly:
`docker run --rm -it -v `pwd`:/app --name sencha ghcr.io/bwbohl/sencha-cmd:latest /bin/bash -c "./build.sh testing; exit"`

What do you think?

riedde:
great idea. I thought something similar. Maybe we can use testing as default if no parameter is set? I have done something similar with the password (for local development use) here: https://github.com/Baumann-Digital/baudi-docker/blob/main/build-run-docker.sh

peterstadler:
My suggestion would be to not provide this as a shell script but rather add this to the ant build.xml.
Since we already have ant in place I think it'd be beneficial to employ this tool for tying together all the build, test, and run steps

riedde:
My intention was to share my way of working, an this was just some addition to build.sh, but I also think we should try to extent the ant build.xml see #440

riedde:
I converted this to draft until the ant way is explored. I'm intrested in working on that to make a proposal.

roewenstrunk:
> My suggestion would be to not provide this as a shell script but rather add this to the ant build.xml. Since we already have ant in place I think it'd be beneficial to employ this tool for tying together all the build, test, and run steps.
> 
> This also includes current `build.sh` which we (imho) should get rid off.

It is an interesting question which tools we expect developers to use. Benni's Docker image has Ant integrated. It would therefore be possible to develop on the Edirom without having Ant installed locally.

riedde:
Are we talking about the same? I am talking about Benni‘s Sencha-Image not about his Edirom-Image.

So what about a short discussion on this at the next developer’s meeting? Maybe we already have some good Workflows and just have to share or combine them?

roewenstrunk:
> Are we talking about the same? I am talking about Benni‘s Sencha-Image not about his Edirom-Image. So what about a short discussion on this at the next developer’s meeting? Maybe we already have some good Workflows and just have to share or combine them?

I was talking about the Sencha-CMD image, too. There is Ant integrated. So technically you don't need Ant locally installed to build Edirom Online.

roewenstrunk:
> > @krHERO To be discussed with the developer team: We can also integrate this script into the build process.
> 
> I like the idea to integrate this into the build process. [build.sh](https://github.com/Edirom/Edirom-Online/blob/aadb46b7ce62c9523c09d6e393b5d5688654659d/build.sh) could test if sencha and ant are available locally and if not then start the build process using the container.
> 
> Small addition starts the build process in the container directly: `docker run --rm -it -v `pwd`:/app --name sencha ghcr.io/bwbohl/sencha-cmd:latest /bin/bash -c "./build.sh testing; exit"`
> 
> What do you think?

@riedde, if you follow Daniel's suggestion, the docker command can't be integrated into `build.sh` because `build.sh` is called inside the docker container. So you would need another shell script for calling docker.

riedde:
Thanks for the hint! I'm going to investigate some solutions and present them at the next meeting.

daniel-jettka:
Proposing small change to current script
* flag "-d" starts docker container and in there the script is executed in the former way

peterstadler:
@riedde please do not hit the "update branch" button and merge dev into this branch. This will create "ugly" history for it inserts all those unnecessary merge commits. It's better to rebase the branch but beware that you'll need to eventually reset hard your local clone.

riedde:
@peterstadler Okay, I won't use this button again, but rebasing is not my favourit way of working. It cost a lot of time because there are always conflicts that are mostly hard to handle. Especially when there is to much code formatting.

build.sh
```diff
@@ -1,5 +1,21 @@
 #!/bin/sh
 
+while getopts "d" flag; do
+ case $flag in
+   d) # Handle the -d flag
+   # run docker
+   echo "Running in Docker..."
+    # run docker
+    docker run --rm -it -v `pwd`:/app --name sencha ghcr.io/bwbohl/sencha-cmd:latest /bin/bash -c "./build.sh testing; exit"
```
peterstadler, line 9: instead of hard coding the "testing" flag I suggest to read this from the initial command line arguments and pass it on to the Docker command
```suggestion
    shift $((OPTIND-1)) # remove -d from the list of input arguments
    OPTIONS=${@} # set a variable OPTIONS with the remaining input arguments to pass to the build command.
    # run docker
    docker run --rm -it -v `pwd`:/app --name sencha ghcr.io/bwbohl/sencha-cmd:latest ./build.sh $OPTIONS"
```

With that you can also do e.g. `./build.sh -d production`

Additionally, I removed the `exit` which is not needed imho

build.sh
```diff
@@ -1,5 +1,21 @@
...
+   exit
```
peterstadler, line 10: This is not really needed, imho
```suggestion
```

peterstadler (changes requested):

peterstadler:
@daniel-jettka could you have a look at the code suggestions and whether this works for you?

build.sh
daniel-jettka: If this is omitted, then the build process is run locally too (the script just proceeds with the sencha commands in the lower part), or am I wrong?

build.sh
daniel-jettka: With this I get:
```
~/git-clones/Edirom/Edirom-Online$ sh build.sh -d
build.sh: 37: Syntax error: Unterminated quoted string
```
Yes right, the `exit` is not needed because the container is closed when the script finishes.

daniel-jettka:
> @daniel-jettka could you have a look at the code suggestions and whether this works for you?

see above :-)

build.sh
peterstadler: Yes, you're right, this `exit` needs to stay

build.sh
peterstadler: Tested again and it works for me. For better readability I paste the full file:
```sh
#!/bin/sh

while getopts "d" flag; do
 case $flag in
   d) # Handle the -d flag
   # run docker
   echo "Running in Docker..."
    # run docker
    shift $((OPTIND-1)) # remove -d from the list of input arguments
    OPTIONS=${@} # set a variable OPTIONS with the remaining input arguments to pass to the build command
    docker run --rm -it -v `pwd`:/app --name sencha ghcr.io/bwbohl/sencha-cmd:latest ./build.sh $OPTIONS
   exit
   ;;   
   \?)
   # Handle invalid options
   ;;
 esac
done


# cleaning the build dir
sencha ant clean

# building the app
sencha app build $1

# get additional stuff for exist-db
ant build-plus

# build xar
ant xar
```

daniel-jettka: Yes, thank you. It works

peterstadler (approved):
The code is great now but yet another PR with several unnecessary and history polluting merges …

riedde:
> The code is great now but yet another PR with several unnecessary and history polluting merges …

@peterstadler Could you bring this up at the next community meeting? I think everyone is doing update, rebase and merge in his/her own style at the moment.

Issue #279, opened 2022-07-04
Uncaught error if 'start_documents_uri' not defined in prefs


bwbohl:
app.js:1 Uncaught constructor {msg: 'No preference found with this key', key: 'start_documents_uri', level: 'warn', sourceMethod: 'getPreference', sourceClass: 'EdiromOnline.controller.PreferenceController', …}

bwbohl:
This does not result in any malfunction but just doesn’t seem right and bad style.

bwbohl:
![Image](https://github.com/user-attachments/assets/8790a056-9e57-41da-b8a9-7f6d39d6143b)

bwbohl:
Apparently it is sufficient if the prefs file contains an empty entry in the form:

```xml
<entry key="start_documents_uri" value=""/>
```

daniel-jettka:
So this is where it may happen that the warning is thrown if there is no entry in the prefs file for `start_documents_uri`: 
(1) https://github.com/Edirom/Edirom-Online/blob/develop/app/Application.js#L130
(2) https://github.com/Edirom/Edirom-Online/blob/develop/app/Application.js#L183

What do you think @bwbohl - is there something to fix or a better way to handle things is the entry is not there?

bwbohl:
I think if we add a check (probably in in (2)), wheter the return value of `var uris` is empty or null and if so `return;` that would do the job.

daniel-jettka:
Where and when is the error message thrown anyway? :-)

Can there be an additional parameter here? - https://github.com/Edirom/Edirom-Online/blob/develop/app/Application.js#L185

So that
`var uris = me.getController('PreferenceController').getPreference('start_documents_uri');`

becomes
`var uris = me.getController('PreferenceController').getPreference('start_documents_uri', true);`

The definition of getPreference looks like that - https://github.com/Edirom/Edirom-Online/blob/develop/app/controller/PreferenceController.js#L64
`getPreference: function(key, lax) { ... }`

Would test that if I knew how to throw the error...

bwbohl:
If you lad an edirom online and the edition has not defined a value for key `start_documents_uri` in its preferences the error is thrown. You located the right spoot in the JS that throws the error.

https://github.com/Edirom/Edirom-Online/blob/aadb46b7ce62c9523c09d6e393b5d5688654659d/app/controller/PreferenceController.js#L64-L94

to be exact:

[...]

PR #439, opened 2024-10-10
Fix/279 start documents uri not found

daniel-jettka:
When key start_documents_uri is no provided in prefs file, then an error message was thrown to the logs, which can be seen as too verbose. Some conditions were introduced to prevent the error message for this key. Can be extended to more keys if desired. 

Refs #279 

## How Has This Been Tested?
Built and deployed Edirom with EditionExample. No error on console anymore.

## Types of changes

## Checklist

### DIFF
--- a/app/Application.js
+++ b/app/Application.js
@@ -182,7 +182,7 @@ Ext.define('EdiromOnline.Application', {
     
     openStartDocuments: function() {
         var me = this;
-        var uris = me.getController('PreferenceController').getPreference('start_documents_uri');
+        var uris = me.getController('PreferenceController').getPreference('start_documents_uri', true);
         if(uris){
             window.loadLink(uris);
         }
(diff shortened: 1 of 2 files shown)


roewenstrunk (changes requested):
The `getPreference` function has a second parameter `lax`. You could use that parameter in the `openStartDocuements` function. It will return `null` if there is no data. Then check `uris` for being `null`. No `ignoredKeys` in `getPreferences` needed.

daniel-jettka:
> The `getPreference` function has a second parameter `lax`. You could use that parameter in the `openStartDocuements` function. It will return `null` if there is no data. Then check `uris` for being `null`. No `ignoredKeys` in `getPreferences` needed.

Thanks for the addition! I tried that with values true and 1 for parameter lax, and it did not work. Should it get another value?

daniel-jettka:
> The `getPreference` function has a second parameter `lax`. You could use that parameter in the `openStartDocuements` function. It will return `null` if there is no data. Then check `uris` for being `null`. No `ignoredKeys` in `getPreferences` needed.

Ah, wait... I had a caching issue. Will try again.

roewenstrunk (approved):
That looks good to me

bwbohl:
Thanks @daniel-jettka & @roewenstrunk for fixing this ;-)

Issue #366, opened 2024-02-15
Remove unused xql files


bwbohl:
- [x]  getAnnotationParticipants.xql (sameAs #324 )

daniel-jettka:
What's the way to check?

Like this? `find . -name "*.xql" > xqlfiles`

and then for all, e.g.: `grep -R  "pre-install.xql" *`

bwbohl:
could be a feasible solution would like to see the list of deletion candidates you come up with!

daniel-jettka:
Deletion candidates:

- [ ] getAnnotationParticipants.xql (sameAs #324 )
- [ ] loginToTextGrid.xql (sameAs #322 )
- [ ] checkTextGridLogin.xql (sameAs #322 )
- [ ] getAnnotationsInRendering.xql
- [ ] controller.xql

peterstadler:
what about `getRendering.xql`? is this still used?

daniel-jettka:
> What's the way to check?
> 
> Like this? `find . -name "*.xql" > xqlfiles`
> 
> and then for all, e.g.: `grep -R "pre-install.xql" *`

```
find . -name "*.xql" | while read -r line ; do
   echo "*** $line"
   grep -R "${line##*/}" * 
done
```

daniel-jettka:
> what about `getRendering.xql`? is this still used? (I couldn't find any reference to it)

Yes, thank you for checking! Unfortunately I have just missed it, it seems.

PR #442, opened 2024-10-11
Ftr/366 remove unused xql

daniel-jettka:

Removing another unused xql which I missed while cleaning up before

Refs #366

## How Has This Been Tested?

Built, deployed, manual visual check

## Types of changes

- cleanup

## Checklist


roewenstrunk (approved):
I think it is safe to delete the file. The called XSLT is also not present any more. @bwbohl, you did some changes to the file. What do you think?

bwbohl:
The changes I did were only when I updated all XQuery files to have a more consistent formatting. When doing so I did not check whether the files were actually being called. So it’s a go from me.

bwbohl (approved):

Issue #443, opened 2024-10-11
[BUG] debug variable is not declared but referenced


roewenstrunk:
**Describe the problem**

There are some checks if a `debug` variable is not `null` and `true` to generate debug output (see for example https://github.com/Edirom/Edirom-Online/blob/c96595da78b482f87a4fd67f7a75e09062b11dd6/app/view/window/source/SourceView.js#L149). These checks fail if the variable has never been declared. A ReferenceError is thrown.

**To Reproduce**

I have installed the default branch with the data of the clarinet quintet in an exist-db. If you then open the autograph, the error appears.

PR #444, opened 2024-10-11
fix ReferenceError when `debug` is not defined

roewenstrunk:

If no `debug` variable is set, there are some `ReferenceErrors`. The fix checks whether the variable is declared.

Refs #443

## How Has This Been Tested?

I have installed the default branch with the fix and with the data of the clarinet quintet in an exist-db. If you then open the autograph, the error does not appear any more.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have performed a self-review of my code
- [x] I have read the [CONTRIBUTING](https://github.com/Edirom/Edirom-Online/blob/develop/CONTRIBUTING.md) document.

### DIFF
--- a/app/controller/window/source/MeasureBasedView.js
+++ b/app/controller/window/source/MeasureBasedView.js
@@ -157,13 +157,13 @@ Ext.define('EdiromOnline.controller.window.source.MeasureBasedView', {
         var me = this;
         
 
-        if(debug !== null && debug) {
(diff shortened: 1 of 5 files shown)


peterstadler:
`typeof null`    returns "object". Ist das ein Problem?

peterstadler (approved):
Thanks for the f2f explanation @roewenstrunk , makes sense!

Issue #154, opened 2021-12-01
Check english and german language files


roewenstrunk:
Check if the english and german language files have equal entries.

bwbohl:
Schematron could be very useful here.

peterstadler:
It'd probably quite easy to add a test to the GitHub actions to check for this issue (be it schematron or what have you).

On the other hand that would not help customizations of the Edirom where users might alter those dictionaries or even provide their own (I assume this is possible)? So maybe we should even run this check at startup and output a warning if some misalignment was found?

peterstadler:
I'm volunteering to implement such a test but still need some guidance on where and when to run such a test:

* when installing Edirom Online, i.e. as a post-install-script within eXist?
* when building Edirom Online, i.e. as part of the (yet to be implemented) test suite? (This would allow to test this locally as well as on any CI/CD system)

roewenstrunk:
Hi @peterstadler, Ideally, we would have two tests. When the Edirom is built, we can only test the information that is delivered with the Edirom or the information that is built in via a fork. We often also allow project-specific customizations through information in the data packages, but these are not available neither when building nor when installing the Edirom. It might be a good idea to offer tests for data packages here.
I would suggest that we first start with a test when building the Edirom. This can then also be offered stand-alone for data packets. What do you think?

bwbohl:
We could decide on whether to have the schematron for this in the https://github.com/Edirom/edirom-edition-schema repository then it could be used 'idendependently'.

The fist aim could be to just check the lang files when editing them, e.g. by referencing the schematron from the files. Checking this on build seems like a good idea. Additionally we could add a GitHub Action that performs the check when any of the files gets modified in a PR to assure that all languages cover all terms e.g. in case something get added or removed.

For data packages the tricky part is they don’t have to provide full coverage of all language variables but might just provide those that should be overridden. The Edirom Online uses its default language files as fallback in those cases.

PR #454, opened 2024-10-15
add XQuery test to check for missing entries in language files

peterstadler:

This PR introduces a new XQuery based check for missing entries in language files

Refs #154

## How Has This Been Tested?

Checked locally by modifying language files and checking results of running `ant check-language-files`

## Types of changes

## Checklist

### DIFF
--- a/add/data/xslt/i18n/en.xml
+++ b/add/data/xslt/i18n/en.xml
@@ -154,7 +154,7 @@
     <entry xml:id="handDesc">Description of Hands</entry>
     <entry xml:id="handNotes">Notes on hands</entry>
     <entry xml:id="itemList">Item List</entry>
-    <entry xml:id="placeName">Ort</entry>
+    <entry xml:id="placeName">Place name</entry>
     
     <entry xml:id="respStmt">Responsibility</entry>
     <entry xml:id="profileDesc">Non-bibliographic aspects</entry>
--- a/add/data/xslt/i18n/en.xml
+++ b/add/data/xslt/i18n/en.xml
@@ -6,8 +6,27 @@
     <entry xml:id="titleStmt">Title Statement</entry>
     <!-- marc keys : http://www.loc.gov/marc/relators/relaterm.html  -->
     <entry xml:id="cmp">Composer</entry>
+    <entry xml:id="cnd">Conductor</entry>
+    <entry xml:id="csl">Consultant</entry>
+    <entry xml:id="dst">Distributor</entry>
+    <entry xml:id="dtc">Data contributor</entry>
+    <entry xml:id="edt">Editor</entry>
+    <entry xml:id="fnd">Funder</entry>
+    <entry xml:id="led">Lead</entry>
     <entry xml:id="mcp">Music copyist</entry>
+    <entry xml:id="mus">Musiker</entry>
+    <entry xml:id="mrk">Markup Editor</entry>
+    <entry xml:id="pbd">Verlagsleiter</entry>
+    <entry xml:id="pbl">Herausgeber</entry>
+    <entry xml:id="pdr">Project director</entry>
+    <entry xml:id="rce">Tontechnik</entry>
+    <entry xml:id="rth">Research team head</entry>
     <entry xml:id="scr">Scribe</entry>
+    <entry xml:id="spn">Sponsor</entry>
+    <!-- Identifier Klassen -->
+    <entry xml:id="labelCode">Labelcode</entry>
+    <entry xml:id="orderNo">Order number</entry>
+    <entry xml:id="EAN">Article number (EAN)</entry>
     <!-- Titelei -->
     <entry xml:id="publication">Publikation</entry>
     <entry xml:id="series">Series</entry>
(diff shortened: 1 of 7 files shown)


peterstadler:
I've also added the new check "check-language-files" to the Github actions:
1. The run https://github.com/Edirom/Edirom-Online/actions/runs/11352963364/job/31576942488 failed due to a missing entry in the German language file
2. After adding the missing entry the Github action passed: https://github.com/Edirom/Edirom-Online/actions/runs/11353084552/job/31577350564

roewenstrunk:
When I run the test, I get an error because I don't have the right libraries for ant: `[java] Could not find net.sf.saxon.Query. Make sure you have it in your classpath`.
Would you like to write a minimal description in the [wiki](https://github.com/Edirom/Edirom-Online/wiki) of how this type of tests work and what requirements must be met?

peterstadler:
I have added some basic information regarding dependencies in the Readme: https://github.com/Edirom/Edirom-Online/blob/3e57257bc5d1f8d095c715c7c93cfc1282c7af85/README.md?plain=1#L104-L110

peterstadler:
pinging @bwbohl and @roewenstrunk for reviewing

bwbohl:
Ok, I successfully tested the lang-test by doing the following:

1. Use [bwbohl/sencha-cmd](https://github.com/bwbohl/sencha-cmd/pkgs/container/sencha-cmd) as described in the [README](https://github.com/Edirom/Edirom-Online?tab=readme-ov-file#building-locally).

2. Install missing dependecies for the tests

   ```
   apt-get update
   ```
 
   ```
   apt-get install libsaxonhe-java
   ```
   
 3. from within '/app' run the test, referencing the JARs and the buildfile
 
 ```
 ant -lib /usr/share/java/ -buildfile testing/ant-tests.xml check-language-files
 ```
 
This reported nothing and when changing the name of one entry it reported the original term missing in the modified file and the modified term in the other file. Everything as expected, very nice!

bwbohl (approved):
Just a short follow-up: Should we also check `add/data/xslt/iq8n`?

bwbohl:
Thanks for imrproving!
Lot’s of discrepancies surfaced this way. Probably we should fix. them before merging ;-)

bwbohl:
Just trying to fix the files and realised maybe we should also cover duplicates in the error report. Currently this would be the result for a duplicate entry in one file:
```xml
<key missing="" available="en de">csl</key>
```

peterstadler:
Regarding duplicates: Shall I add proper output for that case or is the current (undocumented) way of surfacing duplicates ok for now?

bwbohl:
I’d prefer making it clean now, so e.g.

```xml
<key missing="" available="en de" duplicate="de">csl</key>
```

add/data/xslt/i18n/en.xml
```diff
@@ -119,25 +152,32 @@
     <entry xml:id="settlement">Settlement</entry>
     <entry xml:id="supportDesc">Support Description</entry>
     <entry xml:id="handDesc">Description of Hands</entry>
+    <entry xml:id="handNotes">Notes on hands</entry>
     <entry xml:id="itemList">Item List</entry>
-    <entry xml:id="pubStmt">Publication Statement</entry>
+    <entry xml:id="placeName">Ort</entry>
```
peterstadler, line 157: Really German? Or should that rather be an English term?

add/data/xslt/i18n/en.xml
```diff
@@ -119,25 +152,32 @@
...
+    
+    <entry xml:id="title_desc">Descriptive title</entry>
+    <entry xml:id="title_main">Main title</entry>
+    <entry xml:id="title_sub">Subordinate title</entry>
+    
     <!-- FreiDi /content -->
     <entry xml:id="width">width</entry>
     <entry xml:id="height">height</entry>
     <entry xml:id="descLink">Source Description</entry>
+    <entry xml:id="singers">Sänger</entry>
```
peterstadler, line 181: Really German? Or should that rather be an English term?

peterstadler:
added explicit warning for duplicate keys. Output looks like:
```xml
<results>
   <key duplicates="de en">controller.window.Window_renderingView</key>
   <key missing="en" available="de">global_execute</key>
   <key duplicates="de">controller.window.Window_summaryView</key>
   <key missing="de" available="en">global_execue</key>
</results>
```

add/data/xslt/i18n/en.xml
bwbohl:
```suggestion
    <entry xml:id="singers">Singers</entry>
```

add/data/xslt/i18n/en.xml
bwbohl:
```suggestion
    <entry xml:id="placeName">Place name</entry>
```

bwbohl:
After retesting I have to report that it works fine for the files in `locale` but does not report duplicates in the new style when there are some in the ì18n` files.

Moreover if there˚s a duplicate in one of the `locale` files and another in the ì18n` files only the `locale` gets reported.

testing/check-language-files.xq
```diff
@@ -0,0 +1,56 @@
...
+        where count($entry) ne count($language-files)
+        let $available :=
+        (
+            $entry/ancestor::langFile/lang ! string(.), (: language files in add/data/locale :)
+            $entry/parent::language/@xml:lang ! string(.) (: language files in add/data/xslt/i18n :)
+        )
+        let $missing := $languages[not(. = $available)]
+        let $duplicates :=
+            for $i in $entry
+            group by $langkey := $i/ancestor::langFile/lang
```
bwbohl, line 43: fix duplicates test for ì18n` files

```suggestion
            group by $langkey := ($i/ancestor::langFile/lang, $i/parent::language/@xml:lang)
```

testing/ant-tests.xml
```diff
@@ -48,5 +48,43 @@
...
+                Missing entries in language files at /add/data/xslt/i18n: &#10;${xquery.result}
+                <condition>
+                    <!--
+                        If all goes well, "<results/>" will be returned of string-length=10
+                        Any errors reported will add to this string-length hence cause the test to fail
+                    -->
+                    <length string="${xquery.result}" when="greater" length="10"/>
+                </condition>
+            </fail>
+        </sequential>
```
bwbohl, line 87: Execute both tests in parallel to make sure both report their output

```suggestion
        <parallel>
            <sequential>
                <local name="xquery.result"/>
                <java classname="net.sf.saxon.Query" failonerror="false" outputproperty="xquery.result">
                    <arg value="testing/check-language-files.xq"/>
                    <arg value="{http://www.edirom.de/ns/1.3}language-files-path=../add/data/locale?select=*.xml"/>
                </java>
                <fail>
                    Missing entries in language files at /add/data/locale: &#10;${xquery.result}
                    <condition>
                        <!-- 
                            If all goes well, "<results/>" will be returned of string-length=10
                            Any errors reported will add to this string-length hence cause the test to fail
                        -->
                        <length string="${xquery.result}" when="greater" length="10"/>
                    </condition>
                </fail>
            </sequential>
            <sequential>
                <local name="xquery.result"/>
                <java classname="net.sf.saxon.Query" failonerror="false" outputproperty="xquery.result">
                    <arg value="testing/check-language-files.xq"/>
                    <arg value="{http://www.edirom.de/ns/1.3}language-files-path=../add/data/xslt/i18n?select=*.xml"/>
                </java>
                <fail>
                    Missing entries in language files at /add/data/xslt/i18n: &#10;${xquery.result}
                    <condition>
                        <!--
                            If all goes well, "<results/>" will be returned of string-length=10
                            Any errors reported will add to this string-length hence cause the test to fail
                        -->
                        <length string="${xquery.result}" when="greater" length="10"/>
                    </condition>
                </fail>
            </sequential>
        </parallel>
```

peterstadler:
Thank you very much @bwbohl for the thorough review! 
In added your suggestions (great idea to run the tests in parallel!) and finally did some very explicit typing ;) 

Works on my machine but it'd be great if you could have yet another (last) look.

bwbohl (approved):
Added some minor code linting.
Works like a charm, great implementation work, love it!

PR #461, opened 2024-11-07
Fix/441 Fail if no language is set

### LINKED ISSUE #441: [BUG] Edirom fails when no edition lang is defined
**Describe the problem**
When no default lang is defined in the data the edirom fails. I think the problem occurs here https://github.com/Edirom/Edirom-Online/blob/8742faa6240acbd6fd05c58cbe9b1c1f7ef02eec/add/data/xqm/edition.xqm#L76

**To Reproduce**
Delete the language from the preferences and/or the edirom-edition-file.

**Expected behavior**
If no lang is defined the edirom should use the default lang of the software (https://github.com/Edirom/Edirom-Online/blob/8742faa6240acbd6fd05c58cbe9b1c1f7ef02eec/add/data/prefs/edirom-prefs.xml#L25). I remember, that we already fixed this but maybe not for this case.

**Data**
Use any data and play with the lang definition.

**Edirom information**
 - Version on develop (2024-10-11)

**Environment information**
- Standard

riedde:

Fixes #441

## Types of changes
- 
## Checklist


riedde:
I forgot something!

add/controller.xql
```diff
@@ -35,8 +35,8 @@ return
         (: forward index.html to index.xql :)
         <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
             <forward url="index.xql">
-                <set-header name="Set-Cookie" value="edirom-language={$langVal}" />
```
bwbohl, line 38: Why do you remove setting the cookie?

add/data/xqm/util.xqm
```diff
@@ -287,12 +287,19 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
  :)
 declare function eutil:getPreference($key as xs:string, $edition as xs:string?) as xs:string {
 
-    let $preferencesFile := 
+    let $prefFileCustom := 
         try { doc(edition:getPreferencesURI($edition)) }
         catch * { util:log-system-out('Failed to load preferences') }
-
+    
+    let $prefFileDefault := doc($edition:default-prefs-location)
```
bwbohl, line 294: isn't this already handled in the getPreferences.xql`

add/controller.xql
riedde: The cookie is always set and blocks a conditional for getting the language. This is only used in the conditional blocked by the cookie and the cookie is often empty. @nikobeer uses this at RWA online and we think this line of code appears here accidentially (caused by the big merge).

add/data/xqm/util.xqm
riedde: Partially. `getPreferences.xql` only checks if custom prefs are available and if not it calls the default ones. This adjustment allows to lookup a single (required) value if this is not set. The goal is to use custom prefs even if they are not complete defined.

add/data/xqm/util.xqm
```diff
@@ -287,12 +287,19 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
...
+    
+    let $prefFile := if($prefFileCustom//entry/@key = $key) then
+                         $prefFileCustom
+                     else
+                         ($prefFileDefault)
+        
     return
-        $preferencesFile//entry[@key = $key]/@value => string()
+        $prefFile//entry[@key = $key]/@value => string()
```
peterstadler, line 302: This function looks a bit too complicated to me. When there is a custom key available there are at least two lookups that can be saved. What about (pseudo code)
```
let $customKey := do-custom-key-lookup
return
  if ($customKey) then $customKey
  else 
    (: only do default lookup here :)
    try { doc(defaultPrefs)/do-default-lookup } 
    catch ( util:log(…), '')
```

add/data/xqm/util.xqm
```diff
@@ -303,16 +310,10 @@ declare function eutil:getPreference($key as xs:string, $edition as xs:string?)
  : @return The language key
  :)
 declare function eutil:getLanguage($edition as xs:string?) as xs:string {
```
peterstadler, line 312: Methinks the problem still exists that this function `eutil:getLanguage#1` may return the empty string if the function `eutil:getPreference#2` returns the empty string (for whatever reason). 
To prevent this I suggest to simply add a third case with a hard coded value (e.g. "en") as fallback.

add/data/xqm/edition.xqm
```diff
@@ -21,7 +21,7 @@ declare namespace xlink = "http://www.w3.org/1999/xlink";
 
 (: VARIABLE DECLARATIONS =================================================== :)
 
-declare variable $edition:default-prefs-location as xs:string := '../prefs/edirom-prefs.xml';
+declare variable $edition:default-prefs-location as xs:string := request:get-parameter('exist:controller','xmldb:exist:///db/apps/Edirom-Online') || '/data/prefs/edirom-prefs.xml';
```
peterstadler, line 24: I don't think that is an improvement:
a) in most cases XQueries are called directly (i.e. without being called by a controller script), so the hard coded default  will be taken. Hence, if the Edirom Online is installed in a different collection, this will fail.
b) if there is a controller script in any other collection than the root collection, this will also fail, since the value of the `exist:controller` parameter will then point to that collection 
c) there's nothing wrong with a relative path here

add/data/xqm/util.xqm
riedde: Does this work if there are custom keys defined, but one requird is missing? Before the function was just able to use custom or default prefs. I tried to extend the lookup so that custom pref only need to be defined, if they are not the default.

add/data/xqm/util.xqm
riedde: I would not use a hardcoded value here, because this defaulkt is defined in the edirom prefs. We should call it from there, but I agree on doing this as the third case.

add/data/xqm/edition.xqm
riedde: a) why will this fail?
b) okay
c) didn't work before

roewenstrunk: I checked the code and tested the setup without a language set in edition and prefs. The relative path is working and as Peter said the better way.

add/data/xqm/util.xqm
riedde: @peterstadler Okay, I thought about and I think you are right, that the return of an empty value will be a problem, but `eutil:getPreference#2` will not return an empty one. Case: the default is defined there.

add/data/xqm/util.xqm
roewenstrunk: I changed the function here: [046b643](https://github.com/Edirom/Edirom-Online/pull/461/commits/046b643f30765bb997d0d044d4888dc2c4a82da4). Could you please check?

add/data/xqm/util.xqm
roewenstrunk: Does this [68b65ca](https://github.com/Edirom/Edirom-Online/pull/461/commits/68b65caadf1b6423a569e99e2772afb24440eed3) solve the problem? @riedde and @peterstadler

add/data/xqm/util.xqm
```diff
@@ -287,20 +287,20 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
...
-        $prefFile//entry[@key = $key]/@value => string()
-
+        (: If there is a value for the key in the custom preferences file :)
+        if($prefFileCustom//entry/@key = $key) then
+            $prefFileCustom//entry[@key = $key]/@value => string()
+        (: If not, take the value for the key in the default preferences file :)
+        else
+            try { doc($edition:default-prefs-location)//entry[@key = $key]/@value => string() }
+            (: If the key is not in the default file, then there should be an error :)
+            catch * { util:log-system-out(concat('Failed to find the key `', $key, '` in default preferences file')) }
```
peterstadler, line 303: This `catch` will return an empty sequence. This is ok and shouldn't be casted to `xs:string` but you'll need to adjust the function signature.

add/data/xqm/util.xqm
```diff
@@ -310,10 +310,13 @@ declare function eutil:getPreference($key as xs:string, $edition as xs:string?)
  : @return The language key
  :)
 declare function eutil:getLanguage($edition as xs:string?) as xs:string {
+    
     if (request:get-parameter("lang", "") != "") then
         request:get-parameter("lang", "")
-    else
+    else if(eutil:getPreference('application_language', edition:findEdition($edition)) != '') then
```
peterstadler, line 316: This will technically work but I'd prefer to simply check for existence:
```suggestion
    else if(eutil:getPreference('application_language', edition:findEdition($edition))) then
```

In case of failure `eutil:getPreference` will return an empty sequence.
In case of missing values for keys `eutil:getPreference` will return the empty string.
Both cases will be handled properly by both tests (the original and my suggestion)

riedde: And what is, if the "application_language" is definied, but the value is empty?

add/data/xqm/util.xqm
```diff
@@ -310,10 +310,13 @@ declare function eutil:getPreference($key as xs:string, $edition as xs:string?)
...
         eutil:getPreference('application_language', edition:findEdition($edition))
+    else    
+        'en'
```
riedde, line 319: @roewenstrunk @peterstadler the default lang is `de`. This is what I mentioned before. A hard coded value could differ from the default prefs.

add/data/xqm/util.xqm
riedde: @roewenstrunk I can't start the edirom anymore. Does it work on your machine? I got the error that the parameter can't found
Bildschirmfoto 2024-12-02 um 18 17 23

add/data/xqm/util.xqm
peterstadler: That's my second case: "In case of missing values for keys `eutil:getPreference` will return the empty string"

add/data/xqm/util.xqm
riedde: Problem doesn't occure anymore.

add/data/xqm/util.xqm
riedde: Okay, I'll change this to "de" and will be fine with it. But I still can't understand why the software default should be missing some day.
```suggestion
        'de'
```

add/data/xqm/util.xqm
```diff
@@ -287,13 +287,20 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
  :)
 declare function eutil:getPreference($key as xs:string, $edition as xs:string?) as xs:string {
```
riedde, line 288:
```suggestion
declare function eutil:getPreference($key as xs:string, $edition as xs:string?) as xs:string? {
```

add/data/xqm/util.xqm
riedde: @peterstadler Does this https://github.com/Edirom/Edirom-Online/pull/461/files#r1870106894 fit?

peterstadler: Exactly 👍

add/data/xqm/util.xqm
peterstadler: looks good to me!

riedde:
So, if I got it right, all request and discussions are resolved.

add/controller.xql
```diff
@@ -35,8 +35,8 @@ return
...
                 <add-parameter name="lang" value="{$langVal}"/>
+                <add-parameter name="exist:controller" value="{$exist:controller}"/>
```
peterstadler, line 39: what is the purpose of this change? It might be my ignorance but I'm failing to find any reference to the `exist:controller` param in the other changes?

peterstadler:
Thanks for the quick update @riedde . 
Now there's only the conflict in `add/data/xqm/util.xqm` that needs to be resolved before we can merge this PR.

riedde:
Strange. I checked out the branch and merged develop into it locally and there was no conflict. Maybe my git client was able to resolve it when fast-forwarding.

peterstadler (approved):

PR #462, opened 2024-11-07
Fix: Rename util.xqm to eutil.xqm

riedde:
After we decided multiple times to fix the name of the module for consistency reasons, I fixed it.


bwbohl:
I generally approve but think this shoud wait until other PRs are resolved

peterstadler:
I did not take part in the discussion but is it right that the namespace stays "http://www.edirom.de/xquery/util" rather than "http://www.edirom.de/xquery/eutil"?

riedde:
I would prefer to also update the namespace. Does this has further consequences? I think this module is only used inside edirom, so the change would be as breaking as renaming the module, but not more.

riedde:
@daniel-jettka Done, Thanks for your comment!

add/controller.xql
```diff
@@ -3,7 +3,7 @@ xquery version "3.1";
 (: IMPORTS ================================================================= :)
 
 import module namespace edition="http://www.edirom.de/xquery/edition" at "data/xqm/edition.xqm";
-import module namespace eutil = "http://www.edirom.de/xquery/util" at "data/xqm/util.xqm";
+import module namespace eutil = "http://www.edirom.de/xquery/eutil" at "data/xqm/util.xqm";
```
daniel-jettka, line 6: should this also be data/xqm/eutil.xqm ?

testing/XQSuite/eutil-tests.xqm
```diff
@@ -2,7 +2,7 @@ xquery version "3.1";
 
 module namespace eut = "http://www.edirom.de/xquery/xqsuite/eutil-tests";
 
-import module namespace eutil = "http://www.edirom.de/xquery/util" at "xmldb:exist:///db/apps/Edirom-Online/data/xqm/util.xqm";
+import module namespace eutil = "http://www.edirom.de/xquery/eutil" at "xmldb:exist:///db/apps/Edirom-Online/data/xqm/util.xqm";
```
daniel-jettka, line 5: should this also be .../data/xqm/eutil.xqm ?

daniel-jettka (changes requested):
see inline comments

testing/XQSuite/eutil-tests.xqm
riedde: I have never been to this place before, but yes, It won't work this way
```suggestion
import module namespace eutil = "http://www.edirom.de/xquery/eutil" at "xmldb:exist:///db/apps/Edirom-Online/data/xqm/eutil.xqm";
```

add/controller.xql
riedde, line 6:
```suggestion
import module namespace eutil = "http://www.edirom.de/xquery/eutil" at "data/xqm/eutil.xqm";
```

daniel-jettka (approved):

daniel-jettka (approved):

hizclick (approved):
Tested on ExampleEdition

Issue #451, opened 2024-10-15
[BUG] CSS regression


peterstadler:
## Describe the problem

When reviewing the PR #445 with the clarinet quintet data it occurred to me that some CSS had changed. At first I blamed this PR but then I realized that it must have been introduced at some earlier stage since #444 also showed this behavior. 

## To Reproduce
1. build Edirom Online from dev
2. load clarinet data from pipeline build at https://git.uni-paderborn.de/wega/klarinettenquintett-edirom
3. load both xar packages into a fresh existdb
4. launch Edirom Online
5. open "Editionsbericht"

### Expected behavior

running with v1.0.0-beta.5
![Bildschirmfoto 2024-10-15 um 10 10 17](https://github.com/user-attachments/assets/ac532782-da12-4310-8ed7-bf8026a1626c)

### how it looks like 

running with current dev
![Bildschirmfoto 2024-10-15 um 10 11 29](https://github.com/user-attachments/assets/cb5e1f03-7540-4db0-8485-50d527bc7b1e)

bwbohl:
maybe it’s connected to the update of the TEI Stylesheets (#371 )?

bwbohl:
Unfortunately  I cannot reproduce this at the moment, all looks fine in my instance…

bwbohl:
Meanwhile I can confirm this, too…

We should do a proper cleanup of `resources/css/todo.css`
Having it at all was a bad idea in the first place; it came into existence for quick css development but the rules defines should always be properly incorporated in the SASS definitions. Moreover, with the `additional_css_path` preference property it is easy to temporarily add any css file in the database for such purposes.

peterstadler:
There's a similar CSS issue in the Search window, see https://github.com/Edirom/Edirom-Online/pull/463#issuecomment-2465041218

daniel-jettka:
I tried to fix the indentation. @peterstadler Could you check? I mean the heading looks different too... Is this a bigger thing or enough for now to change indentation?

peterstadler:
I think we need to investigate a bit more regarding the root cause of this change. I just compared the CSS files from the [old](https://klarinettenquintett.weber-gesamtausgabe.de/index.html) and the current version:  the old one loads a file `/apps/weber-klarinettenquintett-eol-emeritus/resources/css/main.css`, i.e. from the data collection, not from the Edirom Online collection. Hence, it seems to me that the loading of project specific CSS files fails somehow?!

[...]

PR #463, opened 2024-11-07
156 searchwindow use edition path for collection function

hizclick:

Refs #

## Types of changes

## Checklist

- I have updated the documentation accordingly.

### DIFF
--- a/add/data/xql/search.xql
+++ b/add/data/xql/search.xql
@@ -25,11 +25,16 @@ declare option output:media-type "text/html";
 
 (: FUNCTION DECLARATIONS =================================================== :)
 
-declare function local:filter($node as node(), $mode as xs:string) as xs:string? {
+(:~
+ : Callback function for `kwic:get-summary`
+ : Contrary to the documentation at https://exist-db.org/exist/apps/doc/kwic (1Q18)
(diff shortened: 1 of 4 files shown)


peterstadler:
This still needs some work. There's inline function documentation missing and the kwic results are not output properly, i.e. only the search terms without context.
Additionally, the text input in the upper right does not trigger a search but only opens the search window. That should be fixed, too.

I will look into these issues.

peterstadler:
I think I fixed the remaining points (function documentation, KWIC, top bar text input).

There's yet another thing, i.e. the CSS styling which seems to be missing. Neither the KWIC display seems adequate nor are the search terms highlighted in the text (although the className `searchResult` is injected correctly). This seems related to #451 so I wonder if @bwbohl can reproduce the issue here?

Some Screenshots to illustrate the CSS issue: 

Bildschirmfoto 2024-11-08 um 16 25 31
Bildschirmfoto 2024-11-08 um 16 26 51

peterstadler:
pinging @roewenstrunk and @bwbohl  to review

roewenstrunk (approved):
Tested with Klarinettenquintett

Issue #322, opened 2023-05-31
Remove TextGrid functionality


DennisFriedl:
For example loginToTextGrid.xql and checkTextGridLogin.xql

peterstadler:
there are also references to textgrid here https://github.com/Edirom/Edirom-Online/blob/52a8c2f81b19f12648cc5263388e094ed441357d/add/data/xqm/util.xqm#L138-L149 and here https://github.com/Edirom/Edirom-Online/blob/52a8c2f81b19f12648cc5263388e094ed441357d/app/controller/LinkController.js#L119-L120

I think this should also be removed?

bwbohl:
although d44dfd080d82da253d211cc3d9480f720fa9c52e removed the xql files the code references above probably should be revisited

bwbohl:
@DennisFriedl can you take care of this?

PR #467, opened 2024-11-08
Getting rid of Textgrid

DennisFriedl:
- Removed the function "eutil:getDoc" which is not necessary anymore without Textgrid
- Changed all "eutil:getDoc" calls to standard "doc" calls
- Removed Textgrid case in an OR statement

#322

## How Has This Been Tested?
Tested with the EditionExample.

## Types of changes
- Cleaning up

## Overview

### DIFF
--- a/add/data/xql/getAudioPlayer.xql
+++ b/add/data/xql/getAudioPlayer.xql
@@ -28,7 +28,7 @@ let $docUri :=
         (substring-before($uri, '#'))
     else
         ($uri)
-let $doc := doc($docUri)
+let $doc := eutil:getDoc($docUri)
 let $artist := $doc//mei:titleStmt/mei:respStmt/mei:persName[@role = 'artist']
 let $album := $doc//mei:meiHead/mei:fileDesc/mei:sourceDesc/mei:source[1]/mei:titleStmt/mei:title[1]/text()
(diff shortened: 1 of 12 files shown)


peterstadler:
My take on that would be a little bit different. I actually like the concept of a dedicated `eutil:getDoc` function because it provides a canonical way of getting a document  (from whatever location). Now we are removing support for TextGrid but maybe we'll add support for GitHub in the future?

Additionally, the standard `doc` function is not very robust and will fail with an error if it can't parse the resource as XML. Hence such a wrapper function is a very good place for adding proper error handling.

DennisFriedl:
You are right. Was also thinking of keeping it but thought it would look a little sad with just a return of "doc". But with correct Error handling it gets relevant again.

DennisFriedl:
I reverted the unwanted changes and got only rid of the textgrid part in the wrapper function. @peterstadler what would be a good way of error handling here? The calls of the function expect a document, so shouldn't "doc()" fail when it can't parse?

peterstadler:
It's a little bit beyond the scope of #322 but I couldn't resist to provide my take on a canonical function (including tests) for loading documents within the Edirom.

peterstadler (approved):
tested briefly with the clarinet quintet and all seemed to work

Issue #468, opened 2024-11-08
[BUG] Missing Semicolon in getMeasurePage.xql


DennisFriedl:
**Describe the problem**
Missing semicolon leads to error in measure view

**To Reproduce**
Go to measure view

PR #469, opened 2024-11-08
added missing Semicolon 

DennisFriedl:
Missing semicolon lead to error in measure view. Bug was introduced in #457

Refs #468

## How Has This Been Tested?
With editionExample


peterstadler (approved):
Thanks for the fix and sorry for the bad review last time

Issue #209, opened 2022-03-29
reactivate help


bwbohl:
There used to be a help button in Desktop.Taskbar which is easily reactivated. Nevertheless the help loaded has some issues:

* clicking a hyperlink in the help always opens a new browser tab without actually loading the help
* images do not load

bwbohl:
TaskBar.js line 76

bwbohl:
might be connected to #232 if we e.g. decide to move help there

Issue #24, opened 2013-08-01
View HelpWindow: internal links in help modify browser window.location


bwbohl:
Navigating the help with internal links alters the address in the browsers address line. to e.g. http://localhost:8080/exist/#helpWindow-1059view.window.source.SourceView_viewMenu

bwbohl:
could you please help locating the error?

krHERO:
@bwbohl is this still an issue?

bwbohl:
at least in https://edition.freischuetz-digital.de/
in https://klarinettenquintett.weber-gesamtausgabe.de/ which probably is the latest deployment the help button is actually missing completely

bwbohl:
blocked by #209

krHERO:
#27 is connected to this one?

roewenstrunk:
Hi @riedde , for what reason do you introduced the `<base target="_blank"/>` in https://github.com/Edirom/Edirom-Online/blob/fc3baf87c49e71a1207f6039ad29bdc80af325bc/add/index.xql#L86
This prevents the internal links in text files (so the help also) to work. I know, that we need a better mechanism which does not alter the browser address line, but for the moment…

riedde:
@roewenstrunk I didn't really introduced this, I took it from the old index.html, I guess. Have you testes if the edirom starts without this element?

bwbohl:
This apparently is due to the default `_self` that’s being applied to anchor elements. I’m currently working on this

PR #470, opened 2024-11-15
reactivated help button and fixed xquery parameters problems

daniel-jettka:

The help button and page was commented out, and needed some change to be displayed and trigger help window opening.

Refs #209 

## How Has This Been Tested?

checked manually in Klarinettenquintett and EditionExample

## Types of changes

- Improvement
- Refactoring

## Overview

### DIFF
--- a/add/data/xql/getHelp.xql
+++ b/add/data/xql/getHelp.xql
@@ -45,6 +45,7 @@ let $doc :=
             <param name="tocDepth" value="1"/>
             <param name="graphicsPrefix" value="help/"/>
             <param name="contextPath" value="{$contextPath}"/>
+            (: == passing empty value for docUri (XSLT expects xs:anyURI, but ExtJS view does not provide value) -> github#480 == :)
             <param name="docUri" value="''"/>
         </parameters>
     )
--- a/add/data/xql/getHelp.xql
+++ b/add/data/xql/getHelp.xql
@@ -21,7 +21,6 @@ declare option output:omit-xml-declaration "yes";
 
 let $lang := request:get-parameter('lang', 'en')
 let $idPrefix := request:get-parameter('idPrefix', '')
-let $uri := request:get-parameter('uri', '')
 let $contextPath := request:get-context-path()
 
 let $base := replace(system:get-module-load-path(), 'embedded-eXist-server', '') (:TODO:)
@@ -46,7 +45,7 @@ let $doc :=
             <param name="tocDepth" value="1"/>
             <param name="graphicsPrefix" value="help/"/>
             <param name="contextPath" value="{$contextPath}"/>
-            <param name="docUri" value="{$uri}"/>
+            <param name="docUri" value="''"/>
         </parameters>
     )
 
--- a/add/data/xql/getHelp.xql
+++ b/add/data/xql/getHelp.xql
@@ -21,6 +21,8 @@ declare option output:omit-xml-declaration "yes";
 
 let $lang := request:get-parameter('lang', 'en')
 let $idPrefix := request:get-parameter('idPrefix', '')
+let $uri := request:get-parameter('uri', '')
+let $contextPath := request:get-context-path()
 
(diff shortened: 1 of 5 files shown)


add/data/xql/getHelp.xql
```diff
@@ -21,6 +21,8 @@ declare option output:omit-xml-declaration "yes";
 
 let $lang := request:get-parameter('lang', 'en')
 let $idPrefix := request:get-parameter('idPrefix', '')
+let $uri := request:get-parameter('uri', '')
```
peterstadler, line 24: the uri parameter is passed as an empty string here:
Bildschirmfoto 2024-11-20 um 09 05 03

add/data/xql/getHelp.xql
```diff
@@ -35,14 +37,16 @@ let $doc :=
...
+let $xsl := doc('../xslt/tei/profiles/edirom-body/teiBody2HTML.xsl')
 let $doc :=
     transform:transform($doc, $xsl,
         <parameters>
             <param name="base" value="{concat($base, '/../xslt/')}"/>
             <param name="lang" value="{$lang}"/>
             <param name="tocDepth" value="1"/>
             <param name="graphicsPrefix" value="help/"/>
+            <param name="contextPath" value="{$contextPath}"/>
+            <param name="docUri" value="{$uri}"/>
```
peterstadler, line 49: I'm not exactly sure what is needed here by the XSLT Stylesheet but it might be the document URI of the help page. NB, `document-url($doc)` won't work since you're overwriting the `$doc` variable and it's some in-memory node here at line 49.

peterstadler (changes requested):
While the help button works in my test with the clarinet quintet, the internal links within the help page don't work. Clicking on e.g. the link "Vorbereitung" from the help TOC opens a new browser tab with the Edirom Online index page.

daniel-jettka:
> While the help button works in my test with the clarinet quintet, the internal links within the help page don't work. Clicking on e.g. the link "Vorbereitung" from the help TOC opens a new browser tab with the Edirom Online index page.

This is the problem in #24 right?

Can we move the conversation there and merge this one anyway because #209 was only about reactiviating, not the content and we have the #24 ?

add/data/xql/getHelp.xql
daniel-jettka: added another commit that removes the uri GET parameter: https://github.com/Edirom/Edirom-Online/pull/470/commits/652467f05148b1550d09fc7eae56f8dcc7df51be

image

add/data/xql/getHelp.xql
daniel-jettka: set $docUri to empty string for now; XSLT needs a value - https://github.com/Edirom/Edirom-Online/blob/develop/add/data/xslt/tei/profiles/edirom-body/teiBody2HTML.xsl#L13

peterstadler: I understand that you need this to be merged to move on. Yet, as a reviewer I must complain about the "set $docUri to empty string for now". You should at least leave a (that) comment in the code stating that this is a temporary workaround.

krHERO: Hi and thanks for being "streng", I do prefer that instead of "geschummel". Maybe we do not understand your first review comments correctly until now. We thought you do not like it to have an uri that is empty. 
Could you help a little with description which Solution you have in mind so we can try to fix it in this way. 
Thanks a lot :)

daniel-jettka: Thank you both. I would like to understand the $docUri and what's going wrong here. 
@peterstadler Could you please elaborate? I tried to pass the win.uri in ExtJs but I guess that's not what is needed here

daniel-jettka: I have added a comment in another commit and created a separate issue: #480

roewenstrunk (approved):

roewenstrunk (approved):

peterstadler (approved):
I personally dislike all these merge-dev-into-feature-branch commits and prefer rebasing for a clean history. That said, the code changes look ok, many thanks!

Issue #83, opened 2016-03-29
getAnnotationsOnPage.xql could return "wrong" data.


nikobeer:
It should be possible to exactly specify where annotations are gathered from. As I understand annotations are searched everywhere in '/db/contents/'? So if you have multiple MEI files with annotations you will possibly get "wrong" annotations (see l. 78).

Example from OPERA: working with two MEI files for prerelease final correction -> only "the first" (archived) MEI file is taken into account?

Further problem: This seems to be completely independent from getting annotations for the Annotations List and View? Here the correct file is found (specified). As a consequence you could get different annotations in annotation view compared to source view.

Would it be possible to centralize annotation management within Edirom?

bwbohl:
this is true, while the annotation list is populated from the `work.mei` the getAnnotationsOnPage searches for all `mei:annot` in `prefs:edition_path`

bwbohl:
I would suggest to define corresponding `getAnnotations` functions in the annotation module.

bwbohl:
this seems connected to #156 and #240

bwbohl:
@hizclick yet another one for the collection?

hizclick:
@nikobeer, is this issue already resolved or how can we reproduce the issue?

Issue #240, opened 2022-06-06
getAnnotationsInText.xql should search edition collection


bwbohl:
Currently the xql queries

https://github.com/Edirom/Edirom-Online/blob/d31d591500a8f27f47d95912f9a50107125ee334/add/data/xql/getAnnotationsInText.xql#L33

This will not properly return any annotation as the concept for installing edirom editions no longer requires them in `/db/contents`

bwbohl:
Connected to #83 if we go the way to define a standard function for retrieving annotations in the annotation XQuery module.

bwbohl:
This would need a dataset with annotation in texts. Does anyone from @Edirom/edirom-online-developer have one at hand? Otherwise, we should push this to 1.0.1 or 1.1.0.

bwbohl:
@hizkile basically here’s the same as in #156 so maybe you could tackle them in combination ;-)

PR #487, opened 2024-12-02
Resolving issue #83 and 240

hizclick:

This pull request resolves issue #83 and #240

Refs #

## How Has This Been Tested?

Tested on Klarinettenquintet

## Types of changes

## Overview

### DIFF
--- a/add/data/xql/getAnnotationsInText.xql
+++ b/add/data/xql/getAnnotationsInText.xql
@@ -17,6 +17,8 @@ declare namespace output = "http://www.w3.org/2010/xslt-xquery-serialization";
 declare namespace request = "http://exist-db.org/xquery/request";
 declare namespace xlink = "http://www.w3.org/1999/xlink";
 declare namespace xmldb = "http://exist-db.org/xquery/xmldb";
+import module namespace eutil = "http://www.edirom.de/xquery/eutil" at "../xqm/eutil.xqm";
+
 
 (: OPTION DECLARATIONS ===================================================== :)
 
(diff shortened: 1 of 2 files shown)


daniel-jettka (approved):
In the GUI I do not see any difference, and I understand that is what was intended because it cannot be tested in another way at the moment. Annotations are shown in Klarinettenquintett.

image

add/data/xql/getAnnotationsInText.xql
```diff
@@ -29,10 +31,10 @@ declare option output:media-type "application/json";
     @param $elems The elements to check (most likely measures and zones)
     @returns A sequence of annotation elements
 :)
-declare function local:findAnnotations($uri as xs:string) as element(mei:annot)* {
+declare function local:findAnnotations($uri as xs:string, $edition as xs:string) as element(mei:annot)* {
     
     (: TODO: check if annotations hold URIs or IDRefs :)
-    collection('/db/contents')//mei:annot[matches(@plist, $uri)]
+    collection(eutil:getPreference('edition_path', $edition))//mei:annot[matches(@plist, $uri)]
```
peterstadler, line 37: There's a (rather new) function `edition:collection#1` that will exactly do what you're looking for (I believe):
```suggestion
    edition:collection($edition)//mei:annot[matches(@plist, $uri)]
```

add/data/xql/getAnnotationsInText.xql
```diff
@@ -64,10 +66,11 @@ declare function local:getAnnotations($uriSharp as xs:string, $annotations as el
             }
     }
 };
+let $edition := request:get-parameter('edition', '')
```
peterstadler, line 69: That `let` should be on the next line

peterstadler (changes requested):
The changes to `bootstrap.js` do not belong here, or am I missing something?

peterstadler:
btw, how is this supposed to fix #83 ?!? I do not see anything related to that ticket here?

add/data/xql/getAnnotationsInText.xql
hizclick: Can you clarify this a bit?

hizclick:
> btw, how is this supposed to fix #83 ?!? I do not see anything related to that ticket here?



> btw, how is this supposed to fix #83 ?!? I do not see anything related to that ticket here.

Instead of looking at the annotation in the whole database with '/db/contents/', we search for the specific edition inside the function `findAnnotations()`; however, this fix does not look for specific MEI files. 
A few things has to be clearer.

For example, do we have a defined mechanism to specify which exact MEI file (eg. The first MEI file .... ) ?

add/data/xql/getAnnotationsInText.xql
peterstadler: done. thanks!

PR #271, opened 2022-06-23
Update getMovements.xql

bwbohl:
include parts information

Issue #13, opened 2013-07-03
[ftr] Go to… menu in parts


roewenstrunk:
While going to the page displaying the beginning of a movement or number in scores is easy, we need to have an extra layer for parts in part books.

roewenstrunk:
There is an edition where we implemented a sub structure for parts in "Go to" menu. But this doesn't seem to be in the core repo.

feuerbart:
I copied the parts from the OPERA-implementation and was able to make it work. tested with open-faust sources A3 (with parts) and A2 (without parts).
OPERAs code was never finished, i guess. looks like there is an implementation of partwise and movementwise, but the latter was never finished. probably going to rework it the next days.

see: [9617525](https://github.com/feuerbart/Edirom-Online/commit/9617525cc03bf920969cb540b6993df7442d6481)

bwbohl:
You’re correct to assume that the `movementwise` display was never finished. That was because the decision was to use the `partwise` exclusively in OPERA. But making it work would be great. Maybe there should be a prefs entry for that?

feuerbart:
#493 should solve this issue.

krHERO:
should #493 also close #101

feuerbart:
yes, #101 is basically the same issue.

PR #493, opened 2024-12-05
Ftr/parts in go to menu

feuerbart:
* add parts to goto-menu in source view.
* add preference to change order from partwise to movementwise

Refs #13 
Refs #271 

## How Has This Been Tested?

run in local Edirom-Online, tested with open-faust sources A2 (without parts) and A3 (with parts).

## Types of changes
- Improvement

## Overview


add/data/xql/getMovements.xql
```diff
@@ -27,7 +27,14 @@ let $movements as array(*)* :=
         return
             map {
                 'id': $movement/string(@xml:id),
-                'name': $movement/string(@label)
+                'name': $movement/string(@label),
+	            'parts': for $part in $movement//mei:part
```
peterstadler, line 31: two comments here:
1. cosmetic: please adjust the indentation 
2. I think parts should always be an array to facilitate processing on the JS-frontend. While sequences of maps will properly be casted to JSON arrays when serializing as JSON, a single part would be serialized as a JSON object, not an array. I suggest to wrap the `for $part … ` into an `array { …}`.

peterstadler (changes requested):

feuerbart:
fixed the indentation and added the array-wrap.
tested again with open-faust A2, A3 and A4.

add/data/xql/getMovements.xql
peterstadler: Thanks for the prompt fix!

peterstadler (commented):
The code looks good to me but I haven't had the chance to check the display and the promised new functionality. Someone with parts data should check this in a running Edirom.

daniel-jettka:
@riedde Could you have a look at this one?

add/data/xql/getMovements.xql
```diff
@@ -27,7 +27,16 @@ let $movements as array(*)* :=
         return
             map {
                 'id': $movement/string(@xml:id),
-                'name': $movement/string(@label)
+                'name': $movement/string(@label),
+                'parts': array {
+                    for $part in $movement//mei:part
+                    return
+                        map {
+                            'id': $part/string(@xml:id),
```
riedde, line 35: We don't have a Fallback, if there are no parts. In https://klarinettenquintett.weber-gesamtausgabe.de/index.html (ED1) there are no part IDs and the menu is shown but nor working. Maybe hide the menue in that case? The alternative would be to generate IDs for that case.
Using the Baumann-Digital Data, it works (the parts there have IDs).

peterstadler:
btw, is this also related to #101?

feuerbart:
yes, #101 adresses the same feature.

feuerbart:
added a fix to the issue @riedde mentioned: parts will only be shown if they have ids so it wont break if they don't have.

riedde (approved):
Thank you @feuerbart !

peterstadler (approved):

PR #494, opened 2024-12-05
add path to preferences-file

krHERO:
add path to preferences-file

people, that are not familiar with the structure of the edirom-online repo should be provided with a link instead of mentioning only the "preferences"


daniel-jettka (approved):
That looks good. Thank you.

PR #505, opened 2024-12-17
fix getLinkTarget for MEI 5 work titles

### LINKED ISSUE #337: getLinkTarget.xql fails if window title not found for mei:work

bwbohl:

Remove `mei:titleStmt` when determining the title of the work because in MEI 5 `mei:title` is a direct child of `mei:work`
Closes #337 

## How Has This Been Tested?

With a BAZ-GA MEI 5 work.

## Types of changes

## Overview

--> will know when the PR has run its tests ;-)

### DIFF
--- a/add/data/xql/getLinkTarget.xql
+++ b/add/data/xql/getLinkTarget.xql
@@ -194,8 +194,8 @@ declare function local:getWindowTitle($doc as document-node()?, $type as xs:stri
         (eutil:getLocalizedTitle($doc//tei:fileDesc/tei:titleStmt[1], $lang))
     
     (: HTML :)
-    else if ($type = 'html') then
-        ($doc//head/data(title))
+    else if ($type = 'html' and not(functx:all-whitespace($doc//*:head/*:title))) then
+        $doc//*:head/*:title => normalize-space()
     
     else if($type = 'unknown') then
     
--- a/add/data/xql/getLinkTarget.xql
+++ b/add/data/xql/getLinkTarget.xql
@@ -173,11 +173,13 @@ declare function local:getWindowTitle($doc as node()+, $type as xs:string) as xs
     else if ($type = 'recording') then
         (eutil:getLocalizedTitle($doc//mei:fileDesc/mei:titleStmt[1], $lang))
     
-    (: Source / Score :)
+    (: Source / Score  MEI 4 and newer :)
     else if ($type = 'source' and exists($doc//mei:manifestation/mei:titleStmt)) then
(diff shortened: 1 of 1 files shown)


peterstadler:
Only from looking at it (i.e. without testing) that looks ok. Yet I wonder whether the whole conditional is overly complicated and something like `//work//title[1]` for every case would be enough? 
Otherwise it would be very helpful to have some inline documentation that records the motivation for the different cases.

bwbohl:
Ok, I tried to simplify and globalise a bit in the `local:getWindowTitle`. For a later release we definitely have to revisit all object assumption and put those in corresponding XQuery modules, e.g. `work.xqm` also was a `work:isWork` function but the assumptions there are not identical to those in `getLinkTarget.xql`, that’s why I did not use it for now.

@riedde please test with your dataset that has `mei:work` as root elements in your work encodings. I heavily reduced that bit of the function because `$doc//mei:work` is true for all the cases we had before. I introduced a `worhTitleContainer` variable to deal with the MEI versions.

add/data/xql/getLinkTarget.xql
```diff
@@ -179,17 +188,29 @@ declare function local:getWindowTitle($doc as node()+, $type as xs:string) as xs
...
-    
+
     (: Text :)
-    else if (exists($doc/tei:TEI)) then
+    else if ($type = 'text') then
         (eutil:getLocalizedTitle($doc//tei:fileDesc/tei:titleStmt[1], $lang))
     
     (: HTML :)
     else if ($type = 'html') then
         ($doc//head/data(title))
```
peterstadler, line 198: Since this might result in an empty sequence (which is not supported by the function signature) I propose the following change to make sure a string is returned:
```suggestion
    else if ($type = 'html' and not(functx:all-whitespace($doc//*:head/*:title))) then
        $doc//*:head/*:title => normalize-space()
```

This is to assure that 
1.  HTML elements in the empty namespace as well as in the proper HTML namespace are queried
2. If there is no text content in the title element, the conditional is skipped to return the fallback '[No title found!]'
3. otherwise return the normalized title

peterstadler (changes requested):

roewenstrunk (approved):
Rebased locally and tested with Klarinettenquintett and Pintos

peterstadler (approved):

Issue #504, opened 2024-12-16
[BUG] Edirom is unreachable without Cookies


DennisFriedl:
**Describe the problem**
In the current `dev` branch the Edirom is not reachable with the error "Uncaught TypeError: me.preferences is undefined". 

**To Reproduce**
1. Open an Edirom
2. Open the Developer Tools, delete all Cookies of the Edirom
3. Reload

**Context**
- the problem was introduced with [#fd45a49](https://github.com/Edirom/Edirom-Online/commit/fd45a49c328ecc105d2feb14372a1bd3e1d7f684)
- reintroducing `<set-header name="Set-Cookie" value="edirom-language={$langVal}" />` fixes the error but is a cookie necessary?

peterstadler:
I can confirm the issue (just for the record)

Diginaut:
It seems that the language cannot be read from the preferences file if no cookie is set. Presumably this applies to the preferences query in general.

krHERO:
@feuerbart modify function getLanguage in LanguageController.js
thank you

PR #510, opened 2024-12-18
fix/cookie-lang-prefs: add error handling and default return

feuerbart:
Edirom Online failed if no cookie 'edirom-language' was set because pref 'application_language' was called before preferences were set. this commit is a quick fix, which returns default lang 'en' if 'application_language' could not be found.
Refs #504 

## How Has This Been Tested?
tested with current develop of Edirom Online, open-faust dataset and baumann kantate dataset.

## Types of changes

## Overview

### DIFF
--- a/app/controller/LanguageController.js
+++ b/app/controller/LanguageController.js
@@ -77,9 +77,16 @@ Ext.define('EdiromOnline.controller.LanguageController', {
     },
     
     getLanguage: function() {
-        if(window.getCookie('edirom-language') !== '')
-            return window.getCookie('edirom-language');
-        
-        return getPreference('application_language');
+        /* since removing the cookie, this throws an error because 'application_language' is called before it is defined.
+        * see issue #504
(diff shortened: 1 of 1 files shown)


daniel-jettka (approved):
I can confirm that this works. 

Tested it in Chrome and Firefox. With and w/o cookies set.
Though it always throws the error `console.log('"application_language" not found, using "en" as default.\n');` it works.

DennisFriedl (approved):
Works indeed, thank you!

feuerbart:
> Tested it in Chrome and Firefox. With and w/o cookies set. Though it always throws the error `console.log('"application_language" not found, using "en" as default.\n');` it works.

yes, thats the problem. the function is called up to three times (if i remember correctly) before the preferences are loaded. but after the prefs are loaded, it works fine and the correct lang is returned.
i guess its necessary to investigate further in this, but for 1.0.0 its fine.

Issue #502, opened 2024-12-15
[BUG] getParts.xql fails if part-names can't be translated


feuerbart:
**Describe the problem**
When opening measure-based view of a source with parts, `getParts.xql` throws an error:
> <?xml version="1.0" ?><exception><path>/db/apps/Edirom-Online/data/xql/getParts.xql</path><message>exerr:ERROR The actual return type does not match the sequence type declared in the function's signature: eutil:getLanguageString(xs:string, xs:string*, xs:string) xs:string. Expected cardinality: exactly one, got 0. [at line 264, column 41, source: /db/apps/Edirom-Online/data/xqm/eutil.xqm]
> In function:
> 	eutil:getLanguageString(xs:string, xs:string*, xs:string) [227:12:/db/apps/Edirom-Online/data/xqm/eutil.xqm]
> 	eutil:getPartLabel(node(), xs:string) [32:22:/db/apps/Edirom-Online/data/xqm/eutil.xqm]</message></exception>

i was able to track the problem down:
`getParts.xql` will try to translate the names of the parts with the `edirom-lang-(de|en).xml`-file
(e.g. `<entry key="perfMedium.perfRes.voice" value="Voice (singing)"/>`)
by calling `eutil:getPartLabel()`, which calls `eutil:getLanguageString()` to resolve the lang-key from the lang-file.
if the key (in my case `perfMedium.perfRes.Set (3) Violoncello/Contrabbasso`) does not exist, the return-value will be empty and not match the functions return key and an error is thrown.

explicitly checking the the returned value from the lang-file and returning an empty string if nothing was returned, solved the error: adding
`let $string := if(string-length($string) != 0) then ($string) else ("")`
to the end of `eutil:getLanguageString()`

i cant tell if there might be any sideeffects by adding this line. can someone confirm that its safe?

it probably might be good to rethink translations (esp. of part-names) and implement them consequently (goto-menu with parts doesnt make use of translations so far).
Rel #155 

**To Reproduce**
1. Setup Exist with current Edirom-Online develop and open-faust data
2. Open source A3, go to measure-based view
3. Nothing will be shown, because getParts.xql throws error

**Expected behavior**
no error should be thrown and the name of the parts should be displayed

**Data**
open-faust dataset: https://git.uni-paderborn.de/kio/open-faust

**Edirom information**
[...]

PR #302, opened 2022-12-24
Feature: Auto measure label handling

riedde:
Better handling of Measure labels
- joining labels when multiple measures pointing to same zone (avoids labels like `1/9`)
- also works with already joined labels (multiRests)
- introducing diacritical labels (supplied and del)

needs:

- code clean up
- introducing functions (avoid redundancy)

PR #511, opened 2024-12-18
fix/part-names-key: fix translation of part-names if they are not in …

feuerbart:
solve the problem that getParts.xql throws an error if the part names could not be found in lang keys.

Refs #502 
(Refs #302 )

## How Has This Been Tested?
tested with open-faust A3 (with parts), A2 (without parts)

## Types of changes
- Improvement

## Overview

### DIFF
--- a/add/data/xqm/eutil.xqm
+++ b/add/data/xqm/eutil.xqm
@@ -266,17 +266,11 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
     let $base := system:get-module-load-path()
     let $file := eutil:getDoc(concat($base, '/../locale/edirom-lang-', $lang, '.xml'))
     
-    let $string := 
-        if($file//entry[@key = $key]) 
-        then (
-            $file//entry[@key = $key]/string(@value)
-        ) else (
-            'noValueFound'
-        )  
+    let $string := $file//entry[@key = $key]/string(@value)
     let $string := functx:replace-multi($string, for $i in (0 to (count($values) - 1)) return concat('\{',$i,'\}'), $values)
 
     return
-        $string
+        if($string) then $string else 'noValueFound'
 
 };
 
--- a/add/data/xqm/eutil.xqm
+++ b/add/data/xqm/eutil.xqm
@@ -224,7 +224,7 @@ declare function eutil:getPartLabel($measureOrPerfRes as node(), $type as xs:str
     let $dictKey := 'perfMedium.perfRes.' || functx:substring-before-if-contains($perfResLabel,'.')
 
     let $label :=
-        if(eutil:getLanguageString($dictKey, (), $lang)) then
+        if(eutil:getLanguageString($dictKey, (), $lang) != 'noValueFound') then
             (eutil:getLanguageString($dictKey, (), $lang))
         else
             ($perfResLabel)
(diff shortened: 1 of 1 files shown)


bwbohl:
I tried to reproduce the error and apparently it does not happen in eXist 6.0.1, but it does in 6.3.0, I don’t know why that is but looking at the code it seems more plausible that the error surfaces.

bwbohl:
Here are some screenshots:

develop-branch in eXist-db v6.0.1
Screenshot 2024-12-19 at 14 00 06

develop-branch in eXist-db v6.3.0
Screenshot 2024-12-19 at 14 51 49

feuerbart:fix/part-names-key in eXist-db v6.3.0
Screenshot 2024-12-19 at 14 52 19

bwbohl:
Moreover, I’d like to bring to the attention of @Edirom/edirom-online-developer that the fix provided here is not dpecifically about part labels but concerns all retrievals of langugage variables with `eutil:getLanguageString`, i.e.:

* as the function return-type is `xs:string` it has to return something, which at the moment it does not do if there is corresponding entry in the lang-files
* This PR introduces a fallback `noValueFound` for those cases to satisfy the return-type

bwbohl:
Some thoughts for a more compact query:

```xquery
…
return
    if($string) then $string else 'noValueFound'
```

bwbohl:
This is beyond the initial fix idea:
* If I delete the label of an `instrVoice` the result will be an empty string in the GUI if there is no corresponding lang-file entry, although a sound label could also be retrieved from the corresponding `mei:part`.
* We should come up with a consistent way of how such langVar errors surface in the GUI, e.g., in the XSLTs of the edirom  the key gets returned. This is to some degree helpful, because it eases adding the missing keys.

bwbohl (dismissed):

feuerbart:
thanks for investigating @bwbohl !
like the short version.

i just realized that eutil:getLanguageString() does not make use of getLanguageFile.xql (which also loads the edition-specific lang-file) but loads the edirom-lang-file on its own.
but that probably should be another issue for 1.1.0. (including some documentation for custom editiom-lang-files).

bwbohl:
> i just realized that eutil:getLanguageString() does not make use of getLanguageFile.xql (which also loads the edition-specific lang-file) but loads the edirom-lang-file on its own. but that probably should be another issue for 1.1.0. (including some documentation for custom editiom-lang-files).

Good catch!

add/data/xqm/eutil.xqm
```diff
@@ -266,7 +266,13 @@ declare function eutil:getLanguageString($key as xs:string, $values as xs:string
...
     let $file := eutil:getDoc(concat($base, '/../locale/edirom-lang-', $lang, '.xml'))
     
-    let $string := $file//entry[@key = $key]/string(@value)
+    let $string := 
+        if($file//entry[@key = $key]) 
+        then (
+            $file//entry[@key = $key]/string(@value)
+        ) else (
+            'noValueFound'
+        )
```
bwbohl, line 275:
```suggestion
    let $projectFile := doc(edition:getLanguageFileURI($edition, $lang))
    let $projectString := $projectFile//entry[@key = $key]
    let $ediromString := $file//entry[@key = $key]
    let $string := if($projectString) then $projectString else $ediromString
```

bwbohl: combined with the suggested return statement this could work! But I haven’t tested it

feuerbart: if i add 
`let $edition := request:get-parameter('edition', '')`
it kind of does, but i guess not as intended: it will print the right values but it looks like the edition lang-file won't be found. probably i'm missing something?

should we ignore the custom edition lang-file for now (aka just use the suggested return statement) and come back to this for 1.1.0?

feuerbart:
having a look at the baudi dataset (sources/music/cantatas/baudi-01-8a00c113.xml) i realized it probably might be best to not use the @label to translate via the lang-file but the @codedval. 
baudi uses @codedval and not @label, e.g.:
`<perfRes xml:id="part_06" codedval="flute.i"/>`

daniel-jettka:
Fixed with #515

PR #523, opened 2025-01-10
ensure to return a string

peterstadler:
ensure to return a string when $eventualTitles is the empty sequence.
The error surfaced in the current Open-Faust data where there is a link from the navigator to a non-existent file.

## How Has This Been Tested?

Open-Faust data

## Types of changes

- Improvement

## Overview


daniel-jettka:
Thanks @peterstadler Is this what it should look like? Then I can approve.

Before the fix:

image

After the fix:

image

peterstadler:
@daniel-jettka yes, that is the "expected" behavior:  getLinkTarget.xql does not fail and returns a valid response. The error in the data (the missing file) gets obvious.

PR #527, opened 2025-01-14
fix getPreference function

peterstadler:
the getPreference would only check the URL parameter and not the actual preferences for the key "application_language". This resulted in an always-English frontend independent of the preferences set in the prefs file.

The issue was reported by @obertsalome and can be reproduced with the current Pintos data by simply starting the Edirom without any URL parameter. This will show the frontend GUI elements in English although the "application_language" is set to "de".

## Types of changes
- Improvement

## Overview

### DIFF
--- a/app/controller/PreferenceController.js
+++ b/app/controller/PreferenceController.js
@@ -72,7 +72,11 @@ Ext.define('EdiromOnline.controller.PreferenceController', {
 	        var lang = me.getURLParameter("lang");
 	        if(lang) {
 		        return lang;
-	        } else {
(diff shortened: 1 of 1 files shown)


krHERO (approved):
tested pinto data before and after the changes to the code. before: english, afterwards german.
can also confirm in the preferences.xml it is set to `<entry key="application_language" value="de"/>`

bug
Bildschirmfoto 2025-01-15 um 09 55 01

after fix from Peter
Bildschirmfoto 2025-01-15 um 09 59 02

PR #528, opened 2025-01-14
Fix edition details

peterstadler:

The function `edition:details($uri)?name` would return an element node instead of a string value, thus the endpoint `getEdition.xql` returns e.g. `"name" : "<editionName xmlns=\"http://www.edirom.de/ns/1.3\">Pintos-Edirom von Salome Obert</editionName>"` (among other properties).

This PR fixes this to return only the string value and refactors another function to not duplicate code.

## Steps to reproduce 

Check the Ajax request to `getEdition.xql` which is loaded at every star of the Edirom:  
Bildschirmfoto 2025-01-14 um 18 41 29
(The screenshot already displays the fixed version)

## How Has This Been Tested?
With the Pintos data

## Types of changes
- Improvement
- Refactoring

## Overview

### DIFF
--- a/add/data/xqm/edition.xqm
+++ b/add/data/xqm/edition.xqm
@@ -41,7 +41,7 @@ declare function edition:details($uri as xs:string) as map(*) {
         map {
             "id": $edition/string(@xml:id),
             "doc": $uri,
-            "name": $edition/edirom:editionName
+            "name": $edition/edirom:editionName => fn:normalize-space()
         }
 };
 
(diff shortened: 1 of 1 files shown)


krHERO (approved):
thanks for the fix. i tested with pintos before and after your changes to the code. it works fine.

bug
Bildschirmfoto 2025-01-15 um 10 16 34


after the fix
Bildschirmfoto 2025-01-15 um 10 21 10

peterstadler:
Vielen Dank für das review @krHERO !

Issue #460, opened 2024-11-06
[BUG]


AndreasMunzmay:
**Describe the problem**
no annotation buttons appearing in the text edition

**To Reproduce**:
1. Go to https://openedirom.lilienteich.de/exist/apps/Edirom-Online/index.html?edition=edition-LindP&lang=en
2. Open Edition>Text
3. Show Critical Notes

**Data**
https://git.uni-paderborn.de/kio/open-faust/-/tree/main/content/sources?ref_type=heads

**Additional context**
ALL critical notes of the edition are stored in one common file (critical_notes.xml) because there's only one critical apparatus for the edition as a whole. A note can refer to the music view, to the text view, or to both views of the edition - and of course to one, two, three … up to five out of the seven sources.

feuerbart:
edition id changed, correct link is: https://openedirom.lilienteich.de/exist/apps/Edirom-Online/index.html?edition=edition-open-faust&lang=en

feuerbart:
we have some OPERA-legacy regarding the formatting of the text edition, that i disabled for now. so it should be all edirom defaults.

https://openedirom.lilienteich.de/exist/apps/Edirom-Online/index.html?uri=xmldb:exist:///db/apps/edirom/open-faust/content/edition/edition_text.xml
first text annotation is in l. 48 `Trüb’ durch die Fensterscheiben bricht!` (third paragraph)
in the HTML the annotations exist, but css says `display: none;`
also the annotations button in the menu is displayed in dark grey, seems to be disabled. probably text annotations are disabled somewhere globally? (side note: same annotations in music sources work)

peterstadler:
While #532 fixes some issues regarding text annotations it does not address this issue completely.
In addition to @feuerbart 's last comment about the CSS issues there's yet another bug: when checking "show annotations" globally from the bottom menu and then opening the text view it will fail to process the annotations correctly. 

https://github.com/Edirom/Edirom-Online/blob/2ac07d55bec1ff66db3d33a9cff841b5f27239c4/app/view/window/text/TextView.js#L149 `target` will be `null` – most probably the text content is not already loaded and the `getById` fails to find anything. `showAnnotations` would need to wait for the Ajax request to `getText.xql` to succeed first.

feuerbart:
clicking `show/hide annotations` via the menu works mostly as intended (at least i think so):
`getAnnotationsInText.xql` has proper output:

```
[ {
[...]

PR #532, opened 2025-01-16
fix some annotations-in-text issues

peterstadler:

There surfaced two issues when testing with the Open-Faust data which might be connected to #460.
1. JSONification of the `local:getAnnotations` function was not complete. The function returned a map object `plist` as string, causing some Javascript function to fail.
2. `teitext:getLabel#2` would return a sequence of strings (under some circumstances) which was not allowed by the return value of the function signature.

Refs #460

## Types of changes
- Improvement
- Refactoring

## Overview

### DIFF
--- a/add/data/xql/getAnnotationsInText.xql
+++ b/add/data/xql/getAnnotationsInText.xql
@@ -49,13 +49,12 @@ declare function local:getAnnotations($uriSharp as xs:string, $annotations as el
         let $plist as array(*) :=
             array {
                 for $p in tokenize($annotation/@plist, '\s+')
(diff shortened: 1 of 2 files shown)


peterstadler:
For reference: This can be reproduced with the Open-Faust data by clicking on the "Edition -> Text" entry from the navigator and activating the annotations. `getAnnotationsInText.xql` will return data but in the wrong format (string instead of an object) which causes some Javascript to fail (see Console).
Bildschirmfoto 2025-01-16 um 17 32 12

feuerbart (approved):
looks good to me, annots are returned correctly.

PR #534, opened 2025-01-22
Fix annotations in texts

roewenstrunk:

This PR contains fixes for annotation icons in text view

Refs #460 

## How Has This Been Tested?

Tested with Open Faust data

## Types of changes

## Overview

### DIFF
--- a/app/controller/window/text/TextFacsimileSplitView.js
+++ b/app/controller/window/text/TextFacsimileSplitView.js
@@ -145,7 +145,7 @@ Ext.define('EdiromOnline.controller.window.text.TextFacsimileSplitView', {
     onAnnotationsVisibilityChange: function(view, visible) {
         var me = this;
 
-        if(visible)
+        if(visible && view.getActivePage() !== null)
             window.doAJAXRequest('data/xql/getAnnotationsInText.xql',
                 'GET', 
                 {
--- a/app/view/window/text/TextFacsimileSplitView.js
+++ b/app/view/window/text/TextFacsimileSplitView.js
@@ -411,7 +411,7 @@ Ext.define('EdiromOnline.view.window.text.TextFacsimileSplitView', {
     
     getActivePage: function() {
         var me = this;
-        return me.activePage.get('id');
+        return (typeof me.activePage !== 'undefined' && me.activePage !== null?me.activePage.get('id'):null);
     },
 
 	setChapters: function (chapters) {
--- a/app/view/window/text/TextFacsimileSplitView.js
+++ b/app/view/window/text/TextFacsimileSplitView.js
@@ -259,7 +259,9 @@ Ext.define('EdiromOnline.view.window.text.TextFacsimileSplitView', {
         var me = this;
         var annos = Ext.query('#' + me.id + '_textCont span.annotation');
         Ext.Array.each(annos, function(anno) {
-            Ext.get(anno).hide();
+            var a = Ext.get(anno);
+            a.setVisibilityMode(Ext.Element.DISPLAY);
+            a.hide();
         });
     },
 
--- a/app/view/window/text/TextView.js
+++ b/app/view/window/text/TextView.js
@@ -120,7 +120,7 @@ Ext.define('EdiromOnline.view.window.text.TextView', {
         var me = this;
 
         if(me.annotationsLoaded) {
-            var annos = Ext.query('#' + me.id + '_textCont span.annotation');
+            var annos = Ext.query('#' + me.id + '_textCont div.annotation');
             Ext.Array.each(annos, function(anno) {
                 Ext.get(anno).show();
             });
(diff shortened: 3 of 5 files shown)


roewenstrunk:
@feuerbart , I created an issue in Open Faust (see https://git.uni-paderborn.de/kio/open-faust/-/issues/3). You should add some extra CSS to your edition specific file.

peterstadler (changes requested):
Activating the annotations now displays a tooltip symbol and hovering over it now triggers the tooltip to show, but
1. you can't hide the annotation symbol anymore. Both the global as well as the per view "show/hide annotations" do not work 
2. extra vertical space is introduced in the second column 
Bildschirmfoto 2025-01-23 um 17 40 23

roewenstrunk:
The spacing in the Open Faust data is because of CSS issues in the edition, as I wrote earlier.
The "show/hide" of annotations is now fixed.
@peterstadler

feuerbart:
implemented the css, works fine with the OPERA-table-layout, thanks @roewenstrunk 

Bildschirmfoto 2025-01-24 um 10 57 42

but when i switch to Edirom default layout, the annotation icon is placed below the line:

Bildschirmfoto 2025-01-24 um 10 58 19

seems to be connected to `div { display: block; ... }`.
playing around with `display: inline-block;` and defining some `width`, i could make it work, but i'm not sure if that is a good solution.

roewenstrunk:
Hi @feuerbart , I fixed the inline display and some width issues with `visibility: hidden`. Could you please test?
Hi @peterstadler , could you please accept the changes, you requested?

feuerbart:
Bildschirmfoto 2025-01-30 um 18 24 13

still breaking before the annotation icon. the preceding `<div class="l">` has no `block` defined so the browser defines it as `display: block;` what breaks after the element. setting that manually to `display: inline-block;` fixes the icons position but will break the lines layout.
have done some research but not yet found a solution that will not break after the div for other divs but not for spans.

feuerbart:
i just tested all the options for `display` with divs, none is a solution.
but moving the `<span class="note">...` inside the `<div class="l">...` did it. is there a reason why it is not already there?
the onmouseovers won't work like that but that could be fixed, right?

daniel-jettka:
kind ping to @roewenstrunk @peterstadler

daniel-jettka:
Kristin and I tested and on my machine Linux with Firefox and Chrome it looks like this:

image

Should be right like this, correct? @feuerbart @peterstadler And if yes, could you update your review?

feuerbart:
if you deployed it from git, then that is as it should be. but there is a custom transformation implemented (some OPERA-legacy reason...) that uses a <table> as layout (the only issue here ist the slight shift in height).

daniel-jettka:
Hi @feuerbart Thanks for looking into this again!

What exactly do you mean by "if you deployed it from git"? Do you mean the Edirom or the edition,and which branch?

And is your comment missing sth? It seems like there is a broken sentence... :-)

"...) that uses a

as layout ..."

feuerbart:
yeah, sorry, there is something crucial missing in that sentence, probably some brainafk selection of that part and pressing enter without realizing it...

complete and rephrasedit should be that:
if you build the faust-package from git main branch (edirom version: branch `fix-annotations-in-texts`), then it looks like your screenshot. but there is a custom transformation implemented (some OPERA-legacy reason...) that uses a modified `Transformation.xls` (see [edition/edition_text.xml#L1](https://git.uni-paderborn.de/kio/open-faust/-/blob/main/content/edition/edition_text.xml?ref_type=heads#L1)) and modified `xql/getText.xql` (see prefs.xml, [prefs/prefs.xml#L42](https://git.uni-paderborn.de/kio/open-faust/-/blob/main/content/prefs/prefs.xml?ref_type=heads#L42)) which implement a `<table>`-layout instead of the edirom default `<div>`layout.
if you remove them both (aka default settings) it should look like this: https://github.com/Edirom/Edirom-Online/pull/534#issuecomment-2625146782 because of the lines `display: block;` that will break after itself and push the icon to the next line. changing `display` to e.g. `inline-block` would fix that, but mess up the lines. i haven't found a solution for that breaking with `<div>` yet. as i have tested all possible values for `display`, i could not make one work to show the desired result (that is breaking after the lines `<div>` but not if there is a icons `<div>` following). could be that there is no solution for that.

i just removed the custom settings in the openedirom exist: https://openedirom.lilienteich.de/exist/apps/Edirom-Online/index.html?uri=xmldb:exist:///db/apps/edirom/open-faust/content/edition/edition_text.xml
if you enable the annotation icons, you can see the problem (first annotation in l. 48: "Trüb’ durch die Fensterscheiben bricht!").

the OPERA table-layout works quite fine, the only issue here ist the slight shift of the lines height when you enable the icons. see here: https://github.com/Edirom/Edirom-Online/pull/534#issuecomment-2650181338, the space above l. 48 is slightly bigger than the other lines.



note: i just found out what messed up the other comment: the linebreak was a `<table>` that was not parsed as text but rendered as html and as it had no content, it was just blank space...
just edited that comment to be displayed correctly...

peterstadler (approved):
@roewenstrunk convincingly presented the fixes during today's Community meeting

feuerbart (approved):

PR #554, opened 2025-03-11
build: switch to sencha-cmd:2.1.0

bwbohl:
To stabilise the build this PR changes the used tag of ghcr.io/bwbohl/sencha-cmd from `latest` to `2.1`.


daniel-jettka (approved):
Felt so free to fix the tag reference because "2.1" could not be found but "2.1.0" did :-)

bwbohl:
> Felt so free to fix the tag reference because "2.1" could not be found but "2.1.0" did :-)

thanks for reporting… got to check on the tag extraction in sncha-cmd repo

PR #576, opened 2025-07-02
keep name of edition in the system

krHERO:

this way you can load more than one edition


roewenstrunk (approved):
Wonderful code

Issue #590, opened 2025-09-10
[ftr] create docker-image for website


krHERO:
**Is your feature request related to a problem? Please describe.**
the website about edirom-online is still hosted by github as a gh-page

**Describe the solution you'd like**
create a docker-image for the gh-pages website to host it on our own.

bwbohl:
Is there something that doesn’t work as expected with GitHub Pages?

daniel-jettka:
> Wouldn’t that do the trick, too? (Just curious)
> 
> https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site

Unfortunately not, because in our case external IPs are not provided for internal domains.

peterstadler:
GitHub action for building and deploying the Docker image is installed.
The Website is up and running now at https://edirom-online.edirom.de/

PR #602, opened 2025-09-24
adding git workflow for deploying the github page

hizclick:

Add a git workflow for deployment 

Refs to #590 

## How Has This Been Tested?

Tested on Chrome, Safari and Mozila browsers 
## Types of changes

## Overview

### DIFF
--- a/event.json
+++ b/event.json
@@ -1,1 +1,0 @@
-{"ref":"refs/heads/gh-pages","repository":{"full_name":"someone/fork"}}


peterstadler:
I'm not 100% sure but I think you'll need to somehow associate the Image (which you are building via these actions) with some repo. Usually you do this by deriving the image name from the repo, or alternatively you can add respective labels to the Docker image.

peterstadler (changes requested):
Thanks for all the work @hizclick  but this looks overly complicated now. I think you don't need to dynamically generate the image name but I think it's ok to hard code it here (linke here: https://github.com/Edirom/cartographer-app/blob/d1e0bfe85c4d4565f5c37d8e7ee53f78bb0b6b6f/.github/workflows/docker-image.yml#L22-L23) or derive it from other available parameters (`${{ github.repository }}`).

My point was only that before you put in `ghcr.io/edirom/edirom-online-gh-pages` as image name which did not match the repository.  You probably wanted `ghcr.io/edirom/edirom-online:gh-pages`?

Additionally I prefer putting the labels directly into the Dockerfile rather than adding them programmatically through the GitHub action.

hizclick:
> Thanks for all the work @hizclick but this looks overly complicated now. I think you don't need to dynamically generate the image name but I think it's ok to hard code it here (linke here: https://github.com/Edirom/cartographer-app/blob/d1e0bfe85c4d4565f5c37d8e7ee53f78bb0b6b6f/.github/workflows/docker-image.yml#L22-L23) or derive it from other available parameters (`${{ github.repository }}`).
> 
> My point was only that before you put in `ghcr.io/edirom/edirom-online-gh-pages` as image name which did not match the repository. You probably wanted `ghcr.io/edirom/edirom-online:gh-pages`?
> 
> Additionally I prefer putting the labels directly into the Dockerfile rather than adding them programmatically through the GitHub action.

Thanks, Peter 🙏.  updated the workflow so the image is `ghcr.io/edirom/edirom-online` `(tags: gh-pages and ${{ github.sha }}) `and aligned the Keel payload. I also put the OCI labels into the Dockerfile. This should address your point. Could you take another look?

Issue #591, opened 2025-09-10
[ftr] add edirom development project to new issues and PRs


krHERO:
**Is your feature request related to a problem? Please describe.**
sometimes new issues and PRs are not tracked in the project view, because the "edirom development" project is not assigned.

**Describe the solution you'd like**
add "edirom development" project automatically when a new issue or PR is opened in
* EO
* frontend
* backend

daniel-jettka:
So maybe this has an even easier solution... add field "projects" to issue template

Issue #116, opened 2017-09-27
annotationView single forward/backward


bwbohl:
The AnnotationView offers three view modes for an annotations participants:

1. grid
2. single
3. list

Using **digilib** as image server, when switching to single view the backward and forward icons to flip through the facsimiles fails to load. Tries to load from `/exist/apps/EdiromOnline/pix/annotPreviewForward.png`

Using **leaflet** as image server, when switching to single view, loading backwar and forwardPNGs is not even triggered.

bwbohl:
Also with digilib IIIF image-server the icons are not present. You can see that the should be there if you inspect the HTML:

```html
<div class="annotView">
    <div class="previewArea">
        <div class="previewItem widthLimited" id="ext-gen6106" style="width: 100%; height: 100%;">
<!-- HERE -->
            <div class="stepLeft"></div>
            <div class="stepRight"></div>
<!-- HERE -->
            <div class="imgBox" id="ext-gen6107">
                <img src="[OMITTED]" class="previewImg">
                <input type="hidden" class="previewImgData" value="[object Object]" id="ext-gen6108">
            </div>
            <div class="label">null: Takte&nbsp;1&nbsp;bis&nbsp;14</div>
        </div>
    </div>
</div>
```

bwbohl:
Apparently, the div:s are hidden via CSS with `display:none;`. Changing this makes them visible, but they don’t seem to have a functionality implemented.

They are created in 
https://github.com/Edirom/Edirom-Online/blob/78d4d323a5f3c817a41b97bcbc0896e75b1b3d09/app/view/window/AnnotationView.js#L652-L659

Styling is done in 
https://github.com/Edirom/Edirom-Online/blob/78d4d323a5f3c817a41b97bcbc0896e75b1b3d09/packages/eoTheme/sass/etc/annotation.scss#L323-L345

The presence of this functionality doesn’t seem critical to me; consequently, we could also drop it. Opinions @Edirom/edirom-online-developer ?

PR #611, opened 2025-10-06
Update action - add new issues and prs to project

daniel-jettka:

Updates the action with correct version and secret

Refs #591


krHERO (approved):
looks good in combination with #116
curious to see it live

PR #647, opened 2026-03-03
Add branch filter for main in update-citation workflow

daniel-jettka:
This follows the discussion in https://github.com/Edirom/Edirom-Online-Frontend/issues/135
It reduces the automatic updates of the CITATION.cff to be triggered only on the main branch.
This leads to less action runs (and necessary pulls and merges) while still getting the updated infos (date and commit hash) into the software releases.


peterstadler (approved):
looks good to me

PR #664, opened 2026-04-08
Refine release version checklist in template

daniel-jettka:
Updated release version checklist and added steps for updating the description.


krHERO (approved):
thanks for refinement

